PR #7, opened 2016-11-10
Image feature initial implementation

### LINKED ISSUE #1: Initial image feature implementation
Implementation should follow Editor Recommendations: [Image](http://ckeditor.github.io/editor-recommendations/features/image.html).
Image will be the first widget-based feature, and should be a reference implementation for other widget features to follow. 
Topics connected with Image feature and widgets:
- fake selection - discussed here https://github.com/ckeditor/ckeditor5-engine/issues/627,
- D&D - image and other widgets should be draggable,
- nested editables should be supported (to implement for example image caption), 
- things like captioning and alignment should be easily reusable in other widgets implementations.

UX and UI of image feature will be discussed later on.

### LINKED ISSUE #5028: Initial image feature implementation
Implementation should follow Editor Recommendations: [Image](http://ckeditor.github.io/editor-recommendations/features/image.html).
Image will be the first widget-based feature, and should be a reference implementation for other widget features to follow. 
Topics connected with Image feature and widgets:
- fake selection - discussed here https://github.com/ckeditor/ckeditor5-engine/issues/627,
- D&D - image and other widgets should be draggable,
- nested editables should be supported (to implement for example image caption), 
- things like captioning and alignment should be easily reusable in other widgets implementations.

UX and UI of image feature will be discussed later on.

szymonkups:
Fixes ckeditor/ckeditor5#5028. Creates initial implementation of image feature:
- convertes from model to view for data and editing pipeline,
- converter from view to model for data pipeline,
- selection conversion and fake selection handling,
- mouse down events handling and basic styling.


package.json
```diff
@@ -3,9 +3,19 @@
   "version": "0.0.1",
   "description": "Image feature for CKEditor 5.",
   "keywords": [],
-  "dependencies": {},
+  "dependencies": {
```
Reinmar, line 6: core, engine, ui and ui-default makes sense here. What do other deps do here?

src/converters.js
```diff
@@ -0,0 +1,161 @@
...
+		const nodeAfter = modelSelection.anchor.nodeAfter;
+		const nodeBefore = modelSelection.focus.nodeBefore;
+
+		if ( selected && selected.hasClass( WIDGET_SELECTED_CLASS_NAME ) ) {
+			selected.removeClass( WIDGET_SELECTED_CLASS_NAME );
+		}
+
+		// Check if model selection is over image and create fake selection in the view.
+		if ( !modelSelection.isCollapsed && nodeAfter && nodeAfter.name == 'image' && nodeAfter == nodeBefore ) {
+			let fakeSelectionLabel = 'image widget';
```
Reinmar: This needs to use the `t()` function.

src/converters.js
```diff
@@ -0,0 +1,161 @@
...
+
+		const nodeAfter = modelSelection.anchor.nodeAfter;
+		const nodeBefore = modelSelection.focus.nodeBefore;
+
+		if ( selected && selected.hasClass( WIDGET_SELECTED_CLASS_NAME ) ) {
+			selected.removeClass( WIDGET_SELECTED_CLASS_NAME );
+		}
+
+		// Check if model selection is over image and create fake selection in the view.
+		if ( !modelSelection.isCollapsed && nodeAfter && nodeAfter.name == 'image' && nodeAfter == nodeBefore ) {
```
Reinmar: This deserves `selection.isContaining()` (and hence `range.isContaining()`). The function would check whether something is contained within the selection. I'm only unsure whether this should check whether the selection contains only this node or whether this node is somewhere in this selection. Right now we need only the former, but, in fact, a complete converter would differentiate "widget in a selection" and "focused widget". Only the latter needs a fake selection.

src/converters.js
```diff
@@ -0,0 +1,161 @@
...
+		// Consume img element.
+		if ( !consumable.consume( viewImg, { name: true } ) ) {
+			return;
+		}
+
+		// Create model element.
+		const modelImage = new ModelElement( 'image' );
+
+		// Add src if one is present.
+		if ( viewImg.hasAttribute( 'src' ) && consumable.consume( viewImg, { attributes: [ 'src' ] } ) ) {
```
Reinmar: The src and alt must always be present, so this should be a first check that we do at the beginning.

src/converters.js
```diff
@@ -0,0 +1,161 @@
+/**
```
Reinmar, line 1: I'd ask @scofalik whether these converters are ok, cause for me they look overcomplicated. They are super long while I expect that such a function gets some model node and returns a view node or vice versa. In each converter there should be 3 steps:
1. Check whether what you want to convert is consumable.
2. (In V->M) Check whether it is allowed.
3. Return a proper node.

src/imageengine.js
```diff
@@ -0,0 +1,52 @@
...
+		const editingPipeline = editor.editing;
+
+		// Configure schema.
+		document.schema.registerItem( 'image', '$block' );
+		document.schema.allow( { name: 'image', attributes: [ 'src', 'alt' ] } );
+
+		// Build converter from model to view for data pipeline.
+		buildModelConverter().for( dataPipeline.modelToView )
+			.fromElement( 'image' )
+			.toElement( modelToViewImage( true ) );
```
Reinmar: I'd hope that if we use the converter builders then these functions should be even easier than what is in `converters.js`.

src/mousedownobserver.js
```diff
@@ -0,0 +1,40 @@
+/**
```
Reinmar: This could end in the engine as `MouseObserver` (which supports mousedown, move and up).

src/image.js
```diff
@@ -0,0 +1,86 @@
...
+
+	/**
+	 * @inheritDoc
+	 */
+	init() {
+		const viewDocument = this.editor.editing.view;
+
+		// If mouse down is pressed create selection over whole view element.
+		viewDocument.addObserver( MouseDownObserver );
+		this.listenTo( viewDocument, 'mousedown', this._onMouseDown, { context: this } );
```
Reinmar: We don't use `context` option (it will be removed). Use `() => this._onMouseDown()`.

Reinmar: BTW. since `mousedown` is a single name, I'd call it `onMousedown()`.

src/converters.js
```diff
@@ -0,0 +1,161 @@
...
+
+		if ( modelElement.hasAttribute( 'src' ) ) {
+			viewImg.setAttribute( 'src', modelElement.getAttribute( 'src' ) );
+		}
+
+		if ( modelElement.hasAttribute( 'alt' ) ) {
+			viewImg.setAttribute( 'alt', modelElement.getAttribute( 'alt' ) );
+		}
+
+		return isDataPipeline ?
```
Reinmar: If you used a normal `if()` then you wouldn't have to use `ViewContainerElement()` with all its options twice.

Reinmar: And the code would be more readable because you'd know that this is the same element, but widgetised for the editing.

Reinmar: In fact... I'd use `widgetize()` directly in the `Image` feature. There's a single converter which can be used fro two pipeplines, but for one of them it has to be "enhanced". It's a function composition.

src/mousedownobserver.js
```diff
@@ -0,0 +1,40 @@
...
+ * Fired when mouse button is pressed down on one of the editables.
+ *
+ * Introduced by {@link image.MouseDownObserver}.
+ *
+ * Note that this event is not available by default. To make it available {@link image.MouseDownObserver} needs to be added
+ * to {@link engine.view.Document} by a {@link engine.view.Document#addObserver} method.
+ *
+ * @see image.MouseDownObserver
+ * @event engine.view.Document#click
```
Reinmar: Not a click event.

tests/image.js
```diff
@@ -0,0 +1,92 @@
...
+	let editor, document, viewDocument, imageFeature;
+
+	beforeEach( () => {
+		const editorElement = window.document.createElement( 'div' );
+		window.document.body.appendChild( editorElement );
+
+		return ClassicTestEditor.create( editorElement, {
+			features: [ Image, Paragraph ]
+		} )
+			.then( newEditor => {
```
Reinmar: We rather keep the indentation of the `} )` line.

Reinmar (changes requested):

tests/manual/image.js
```diff
@@ -0,0 +1,24 @@
...
+
+import ClassicEditor from '/ckeditor5/editor-classic/classic.js';
+import EnterFeature from '/ckeditor5/enter/enter.js';
+import TypingFeature from '/ckeditor5/typing/typing.js';
+import ParagraphFeature from '/ckeditor5/paragraph/paragraph.js';
+import HeadingFeature from '/ckeditor5/heading/heading.js';
+import ImageFeature from '/ckeditor5/image/image.js';
+
+ClassicEditor.create( document.querySelector( '#editor' ), {
+	features: [ EnterFeature, TypingFeature, ParagraphFeature, HeadingFeature, ImageFeature ],
```
Reinmar: I'm missing undo and clipboard here. Always include those in manual tests :).

theme/theme.scss
```diff
@@ -0,0 +1,17 @@
+figure {
```
Reinmar: Style `.ck-widget` as a widget. Figure or img have nothing to do here.

theme/theme.scss
```diff
@@ -0,0 +1,17 @@
...
+	&.ck-widget.image {
+		text-align: center;
+
+		img {
+			pointer-events: none;
+		}
+
+		&.ck-widget_selected img, &.ck-widget_selected:hover img{
```
Reinmar: Missing space before `{`.

theme/theme.scss
```diff
@@ -0,0 +1,17 @@
+figure {
+	&.ck-widget.image {
+		text-align: center;
+
+		img {
```
Reinmar: Why is it needed?

tests/manual/image.md
```diff
@@ -0,0 +1,8 @@
+@bender-ui: collapsed
+@bender-tags: view
+
+## Image feature
+
+ - Two images with CKEditor logo should be loaded.
```
Reinmar: Please creates lists usign `* XXX`. No space at the beginning of a line.

theme/theme.scss
Reinmar: What I mean is that we'll need "widget styles" and its the widget container what should be styled. So e.g. you shouldn't be styling the `img` below because the widget system won't know about content of widgets.

theme/theme.scss
Reinmar: Besides, image selected in a blurred editor should not have a blue outline, but grey.

package.json
szymonkups: They are moved to `devDependencies` as they are used in tests.

src/widget/widget.js
```diff
@@ -0,0 +1,75 @@
...
+
+		domEventData.preventDefault();
+
+		// Focus editor if is not focused already.
+		if ( !viewDocument.isFocused ) {
+			viewDocument.focus();
+		}
+
+		// Create model selection over widget.
+		const document = editor.document;
```
Reinmar: `modelDocument`

src/widget/widget.js
```diff
@@ -0,0 +1,75 @@
...
+	 * @param {envine.view.observer.DomEventData} domEventData
+	 * @private
+	 */
+	_onMousedown( eventInfo, domEventData ) {
+		let widgetElement = domEventData.target;
+		const editor = this.editor;
+		const viewDocument = editor.editing.view;
+
+		// If target is not a widget element - check if one of the ancestors is.
+		if ( !isWidget( widgetElement ) ) {
```
Reinmar, line 51: This could be included in `findAncestor()` check somehow.

src/utils.js
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Converts given {@link engine.view.Element} to image widget:
+ * * adds {@link engine.view.Element#addCustomProperty custom property} allowing to recognize image widget element,
+ * * calls {@link image.widget.utils.widgetize widgetize}.
+ *
+ * @param {engine.view.Element} viewElement
+ * @returns {engine.view.Element}
+ */
+export function toImageWidget( viewElement ) {
```
Reinmar, line 18: I think I really like this approach. No subclassing. Everything happens by extension (through custom props). And the exposed API is `make()` and `check()`.

src/widget/utils.js
```diff
@@ -0,0 +1,50 @@
...
+ * * adds custom `getFillerOffset` method returning `null`,
+ * * adds `ck-widget` CSS class,
+ * * adds custom property allowing to recognize widget elements by using {@link image.widget.utils.isWidget}.
+ *
+ * @param {engine.view.Element} element
+ * @returns {engine.view.Element} Returns same element.
+ */
+export function widgetize( element ) {
+	element.setAttribute( 'contenteditable', false );
+	element.getFillerOffset = () => null;
```
Reinmar: That `returnNull` function could be defined once.

src/converters.js
```diff
@@ -0,0 +1,108 @@
...
+}
+
+/**
+ * Returns model to view selection converter. This converter is applied after default selection conversion is made.
+ * It creates fake view selection when {@link engine.view.Selection#getSelectedElement} returns instance of image widget.
+ *
+ * @param {Function} t {@link utils.Locale#t Locale#t function} used to translate default fake selection's label.
+ * @returns {Function}
+ */
+export function modelToViewSelection( t ) {
```
Reinmar, line 69: There should be no special selection converter for image widget. It should be general to widgets. So that label must be handled differently.

src/imageengine.js
```diff
@@ -0,0 +1,57 @@
...
+	 * @inheritDoc
+	 */
+	init() {
+		const editor = this.editor;
+		const document = editor.document;
+		const dataPipeline = editor.data;
+		const editingPipeline = editor.editing;
+
+		// Configure schema.
+		document.schema.registerItem( 'image', '$block' );
```
Reinmar: Image is not a block. Block can have inline elements inside. `<image>` is empty so it's a totally new element, which can be used inside root.

src/imageengine.js
```diff
@@ -0,0 +1,57 @@
...
+		document.schema.allow( { name: 'image', attributes: [ 'src', 'alt' ] } );
```
Reinmar: Both are required attributes. How to define them, @scofalik?

src/imageengine.js
Reinmar: Additionally, `'image'` should be added to `schema.objects` set. It's used by the `insertContent` algorithm to make some decisions on how to selected the pasted thing. And it can be used by more algorithms (e.g. the keyboard support).

src/imageengine.js
```diff
@@ -0,0 +1,57 @@
...
+		return [ WidgetEngine ];
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	init() {
+		const editor = this.editor;
+		const document = editor.document;
+		const dataPipeline = editor.data;
```
Reinmar: `data` and `editing`. That's how we name these variables in most features. Plus, I don't like `document` because it's so deeply anchored in our minds that this is the DOM document that it's causing me problems reading the code. We often call it `doc` therefore.

Reinmar (changes requested):
Some minor issues and possible followups.

src/widget/widget.js
szymonkups: We might add `includeNode` option to `findAncestor` similar as in: https://github.com/ckeditor/ckeditor5-engine/blob/master/src/view/node.js#L129.

PR #14, opened 2016-12-01
Custom handling for delete 

### LINKED ISSUE #13: Custom handler for delete
Extracted from https://github.com/ckeditor/ckeditor5-engine/issues/696#issuecomment-263840273.

Use the view events, cause we shouldn't change the `DataController#deleteContent` semantics to "select".

## Deleting before/after object

E.g.:

* `<obj></obj><p>^x</p>` + backspace
* `<obj></obj><p>^</p>` + backspace

1. Select the object.
2. If the initial selection container is empty, remove it.

### LINKED ISSUE #5037: Custom handler for delete
Extracted from https://github.com/ckeditor/ckeditor5-engine/issues/696#issuecomment-263840273.

Use the view events, cause we shouldn't change the `DataController#deleteContent` semantics to "select".

## Deleting before/after object

E.g.:

* `<obj></obj><p>^x</p>` + backspace
* `<obj></obj><p>^</p>` + backspace

1. Select the object.
2. If the initial selection container is empty, remove it.

szymonkups:
Fixes ckeditor/ckeditor5#5037.

### DIFF
--- a/src/widget/widget.js
+++ b/src/widget/widget.js
@@ -7,7 +7,9 @@ import Plugin from '../../core/plugin.js';
 import WidgetEngine from './widgetengine.js';
 import MouseObserver from '../../engine/view/observer/mouseobserver.js';
 import ModelRange from '../../engine/model/range.js';
+import ModelSelection from '../../engine/model/selection.js';
 import { isWidget } from './utils.js';
+import { keyCodes } from '../../utils/keyboard.js';
 
 /**
  * The widget plugin.
@@ -33,14 +35,17 @@ export default class Widget extends Plugin {
 		// If mouse down is pressed on widget - create selection over whole widget.
 		viewDocument.addObserver( MouseObserver );
 		this.listenTo( viewDocument, 'mousedown', ( ...args ) => this._onMousedown( ...args ) );
+
+		// Handle custom keydown behaviour.
+		this.listenTo( viewDocument, 'keydown', ( ...args ) => this._onKeydown( ...args ), { priority: 'high' } );
 	}
 
 	/**
 	 * Handles {@link engine.view.Document#mousedown mousedown} events on widget elements.
 	 *
+	 * @private
 	 * @param {utils.EventInfo} eventInfo
 	 * @param {envine.view.observer.DomEventData} domEventData
-	 * @private
 	 */
 	_onMousedown( eventInfo, domEventData ) {
 		let widgetElement = domEventData.target;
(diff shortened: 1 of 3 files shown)


src/widget/widget.js
```diff
@@ -64,12 +69,71 @@ export default class Widget extends Plugin {
...
+		// Clone current selection to use it as a probe. We must leave default selection as it is so it can return
+		// to its current state after undo.
+		const probe = ModelSelection.createFromSelection( modelSelection );
+		const isForward = ( keyCode == keyCodes.delete );
+
+		dataController.modifySelection( probe, { direction: isForward ? 'forward' : 'backward' } );
+
+		const objectElement = isForward ? probe.focus.nodeBefore : probe.focus.nodeAfter;
+
+		if ( objectElement && modelDocument.schema.objects.has( objectElement.name ) ) {
```
Reinmar: What if it's not an element, but a text? Actually, what is `nodeAfter/Before` return if we're in a middle of a text?

src/widget/widget.js
```diff
@@ -64,12 +69,71 @@ export default class Widget extends Plugin {
...
+			domEventData.preventDefault();
+			eventInfo.stop();
+
+			modelDocument.enqueueChanges( () => {
+				// Remove previous node if empty;
```
Reinmar: Previous element, not a node. And period.

src/widget/widget.js
```diff
@@ -64,12 +69,71 @@ export default class Widget extends Plugin {
...
+				const previousNode = probe.anchor.parent;
+
+				if ( previousNode.isEmpty ) {
```
Reinmar, line 124: We can already handle here the case where the anchor was in a `<blockquote><p>^</p></blockquote>`. In such case, the blockquote will need to be removed as well. If `modifySelection()` doesn't handle this case yet at all (so we're not even able to write a test for it, then please add it to https://github.com/ckeditor/ckeditor5-engine/issues/710.

src/widget/widget.js
```diff
@@ -64,12 +69,71 @@ export default class Widget extends Plugin {
...
+	 *
+	 * @private
+	 * @param {utils.EventInfo} eventInfo
+	 * @param {envine.view.observer.DomEventData} domEventData
+	 */
+	_onKeydown( eventInfo, domEventData  ) {
+		const keyCode = domEventData.keyCode;
+
+		// Handle only delete and backspace.
+		if ( keyCode !== keyCodes.delete && keyCode !== keyCodes.backspace ) {
```
Reinmar, line 94: We'd need a test that this code will handle also delete/backspace with a modifier key (ctrl e.g.). If this code will ever be changed to e.g. `domEventData.keystroke() == keyCodes.delete` then it wouldn't work the same way.

tests/widget/widget.js
```diff
@@ -110,4 +112,85 @@ describe( 'Widget', () => {
 		sinon.assert.notCalled( focusSpy );
 		expect( getModelData( document ) ).to.equal( '[<widget></widget>]' );
 	} );
+
+	describe( 'delete and backspace handling', () => {
```
Reinmar, line 123: There are no tests for many potential false positive cases like what if the widget was selected in the first place and is preceded by another one, what if it's preceded by a paragraph with text or paragraph with an inline object, or what if we have inline object preceded by a text. It'd be easier if the probing mechanism was exposed. E.g. as I commented there – it's unclear for me (cause I don't know the API in fact :D) what's nodeBefore/After behaviour in case of text nodes. What if we have something like: `<p>xx[]<obj></obj></p>` case and we press backspace? The focus will be moved between "x[x]" and what's the `nodeAfter` then?

Reinmar (changes requested):
The implementation requires more tests and perhaps it could handle the nested block case already.

src/widget/widget.js
szymonkups: `nodeAfter/Before` will return `null` if we are in the middle of the text.

Reinmar: And if we're at the beginning of a text node? Let's say – there was just one letter before our initial selection? Then, will we get that text node?

szymonkups: That's true. Should I handle such situation or leave it to be filtered out by schema checking for objects?

szymonkups: After I while I see that checking for `undefined` name might not be a good idea  :smile:

tests/widget/widget.js
szymonkups: > There are no tests for many potential false positive cases like what if the widget was selected in the first place and is preceded by another one, what if it's preceded by a paragraph with text or paragraph with an inline object, or what if we have inline object preceded by a text.

Deleting itself is handled by default delete mechanism I thought that it should be enough to check  only if modified behavior acts well here. But I think it is worth checking here too.

src/widget/widget.js
Reinmar: Exactly ;>

tests/widget/widget.js
Reinmar: There are far more false positive cases than not. And the risk of this code handling too many cases is, in fact, much worse for the editor than not handling some widget's case, because it may break the general editor behaviour. So when adding some extension to already tested mechanism, we must be sure that this extension is very precisely targetted. It's really more important than checking the positive cases.

Reinmar:
I still miss the case where there's just one letter after/before selection, so the `position.nodeAfter/Before` could (perhaps) return it. I have some feeling that it may break (or at least start checking a text node in `schema.objects()`).

tests/widget/widget.js
```diff
@@ -5,27 +5,33 @@
...
 import { setData as setModelData, getData as getModelData } from 'ckeditor5/engine/dev-utils/model.js';
+import { keyCodes } from 'ckeditor5/utils/keyboard.js';
 
 describe( 'Widget', () => {
 	let editor, document, viewDocument;
 
 	beforeEach( () => {
 		return VirtualTestEditor.create( {
-			plugins: [ Widget ]
+			plugins: [ Widget, Typing ]
```
Reinmar: It's a leftover, right?

tests/widget/widget.js
```diff
@@ -110,4 +120,256 @@ describe( 'Widget', () => {
...
+
+		test(
+			'should delete normally on non-collapsed selection',
+			'<widget></widget><paragraph>[f]oo</paragraph>',
+			keyCodes.backspace,
+			'<widget></widget><paragraph>[]oo</paragraph>'
+		);
+
+		test(
+			'shoulddelete normally on non-object elements',
```
Reinmar: typo

src/widget/widget.js
szymonkups: Added comment: https://github.com/ckeditor/ckeditor5-engine/issues/710#issuecomment-264787546

szymonkups:
Merged with master after docs fixes.

tests/manual/image.md
```diff
@@ -3,3 +3,12 @@
 * Two images with CKEditor logo should be loaded.
 * Hovering over image should apply yellow outline.
 * Clicking on image should apply blue outline which should not change when hovering over.
+
+### Deleting
```
Reinmar, line 7: These tests will need to be moved with the widget API, but it's fine to have them here now.

PR #17, opened 2016-12-12
Arrow keys handling

### LINKED ISSUE #15: Custom arrow keys handling
Extracted from: https://github.com/ckeditor/ckeditor5-engine/issues/696#issuecomment-263840273.

# Custom widget handler for arrows

Use the view events (`keydown` or `selectionChange`).

## Object selected

1. Collapse selection in the right direction (or even base on the `view#selectionChange` which currently implements the same behaviour).
2. Use `Document#getNearestSelectionPosition` with the right direction param to find the next possible position.
3. If not position found... rollback?

## Selection next to an object

1. If selection is not collapsed – do nothing.
2. Two options:
  * Use `DataController#extendSelection()` on a selection clone to probe what's next. If selection focus is before/after an object, it means that this object should be selected.
  * Alternatively – use the walker to find a closest object (in the right direction, ofc). Check if position after/before that object is touching the current selection.
3. Select the found object.

Note: The above algorithm (the option 2a) will automatically work with the case when e.g. we selected an object at the beginning of the root and pressed left arrow. The focus will not be moved any further, so on its right side there'll still be that object and we'll reselect it. :tada:

### LINKED ISSUE #5038: Custom arrow keys handling
Extracted from: https://github.com/ckeditor/ckeditor5-engine/issues/696#issuecomment-263840273.

# Custom widget handler for arrows

Use the view events (`keydown` or `selectionChange`).

## Object selected

1. Collapse selection in the right direction (or even base on the `view#selectionChange` which currently implements the same behaviour).
2. Use `Document#getNearestSelectionPosition` with the right direction param to find the next possible position.
3. If not position found... rollback?

## Selection next to an object

1. If selection is not collapsed – do nothing.
2. Two options:
  * Use `DataController#extendSelection()` on a selection clone to probe what's next. If selection focus is before/after an object, it means that this object should be selected.
  * Alternatively – use the walker to find a closest object (in the right direction, ofc). Check if position after/before that object is touching the current selection.
3. Select the found object.

Note: The above algorithm (the option 2a) will automatically work with the case when e.g. we selected an object at the beginning of the root and pressed left arrow. The focus will not be moved any further, so on its right side there'll still be that object and we'll reselect it. :tada:

szymonkups:
Fixes ckeditor/ckeditor5#5038. Please review after closing: https://github.com/ckeditor/ckeditor5-engine/pull/717.

### DIFF
--- a/src/widget/widget.js
+++ b/src/widget/widget.js
@@ -88,18 +88,33 @@ export default class Widget extends Plugin {
 	 * @param {module:engine/view/observer/domeventdata~DomEventData} domEventData
 	 */
 	_onKeydown( eventInfo, domEventData  ) {
-		this._handleBackspaceAndDelete( eventInfo, domEventData );
-		this._handleArrowKeys( eventInfo, domEventData );
-	}
-
-	_handleBackspaceAndDelete( eventInfo, domEventData ) {
 		const keyCode = domEventData.keyCode;
 
-		// Handle only delete and backspace.
-		if ( keyCode !== keyCodes.delete && keyCode !== keyCodes.backspace ) {
-			return;
+		// Handling delete keys.
+		if ( keyCode == keyCodes.delete || keyCode == keyCodes.backspace ) {
+			if ( this._handleDelete( keyCode == keyCodes.delete ) ) {
+				domEventData.preventDefault();
+				eventInfo.stop();
+			}
 		}
 
+		// Handling arrow keys.
+		if ( isArrowKeyCode( keyCode ) ) {
+			if ( this._handleArrowKeys( keyCode == keyCodes.arrowdown || keyCode == keyCodes.arrowright ) ) {
+				domEventData.preventDefault();
+				eventInfo.stop();
+			}
+		}
+	}
+
+	/**
+	 * Handles delete keys.
+	 *
+	 * @private
+	 * @param {Boolean} isForward Set to true if delete was performed in forward direction.
+	 * @returns {Boolean|undefined} Returns `true` if keys were handled correctly.
+	 */
+	_handleDelete( isForward ) {
 		const modelDocument = this.editor.document;
 		const modelSelection = modelDocument.selection;
 
@@ -108,12 +123,9 @@ export default class Widget extends Plugin {
 			return;
 		}
 
-		const objectElement = this._getObjectNextToSelection( keyCode == keyCodes.delete );
+		const objectElement = this._getObjectNextToSelection( isForward );
 
 		if ( objectElement ) {
-			domEventData.preventDefault();
-			eventInfo.stop();
-
 			modelDocument.enqueueChanges( () => {
 				// Remove previous element if empty.
 				const previousNode = modelSelection.anchor.parent;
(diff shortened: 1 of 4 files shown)


src/widget/widget.js
```diff
@@ -140,4 +187,69 @@ export default class Widget extends Plugin {
...
+	}
+}
+
+// Returns the selected element. {@link module:engine/model/element~Element Element} is considered as selected if there is only
+// one range in the selection, and that range contains exactly one element.
+// Returns `null` if there is no selected element.
+//
+// @param {module:engine/model/selection~Selection} modelSelection
+// @returns {module:engine/model/element~Element|null}
+function getSelectedElement( modelSelection ) {
```
szymonkups: I was wondering whether this should stay here or be moved as model's selection method. We have similar one in view's selection.

Reinmar: Makes sense moving it there. But it needs to have the same semantics like the method in the view. Are they identical?

Reinmar:
All works except this case:
dec-13-2016 17-33-06

src/widget/widget.js
szymonkups: Yes, they have same semantics. I will create an issue in [ckeditor5-engine](https://github.com/ckeditor/ckeditor5-engine).

szymonkups:
I've fixed issue from previous comment and added test checking it. But I noticed one thing - when data is loaded and image is the first element in the document filler is added before it. This is probably an engine issue, could you confirm?

Reinmar:
> I've fixed issue from previous comment and added test checking it. But I noticed one thing - when data is loaded and image is the first element in the document filler is added before it. This is probably an engine issue, could you confirm?

Hm... Hard to tell. The `_getDefaltRange()` method is called couple of times when content is being loaded. It's last calls seem to return a range containing the image:

image

So, I think that this is a bug in the engine. I'll report a ticket for it, but please change this sample so image isn't the first element in the root.

Reinmar:
Reported https://github.com/ckeditor/ckeditor5-engine/issues/730.

src/widget/widget.js
```diff
@@ -128,6 +130,51 @@ export default class Widget extends Plugin {
...
+	 * Handles arrow keys.
+	 *
+	 * @param {Boolean} isForward Set to true if arrow key should be handled in forward direction.
+	 * @returns {Boolean|undefined} Returns `true` if keys were handled correctly.
+	 */
+	_handleArrowKeys( isForward ) {
+		const modelDocument = this.editor.document;
+		const schema = modelDocument.schema;
+		const modelSelection = modelDocument.selection;
+		const objectElement = getSelectedElement( modelSelection );
```
Reinmar: Should use the new engine function

src/widget/widget.js
```diff
@@ -128,6 +130,51 @@ export default class Widget extends Plugin {
...
+
+		// if object element is selected.
+		if ( objectElement && schema.objects.has( objectElement.name ) ) {
```
Reinmar, line 151: This code has nothing to do with widgets. This is interesting, cause it means that we can move all that to the engine easily. That can be a good option. Perhaps it will be easily doable to have basic widget constructs in the engine. However, as a what? Addition to the editing controller? Maaaybe. But it's so much unnecessary code that I don't know ;|.

szymonkups: That's a valid point. We can make a discussion about it when we will extracting widgets functionalities from image plugin.

PR #23, opened 2016-12-29
ImageStyle

### LINKED ISSUE #18: Image styles (alignment)
During the talk with @Reinmar about image alignment, it turned out that we could create more general solution to apply any styles (CSS classes) to images, and not only alignment ones.
It should introduce new classes:

* `ImageStyleCommand`
Command that will apply proper style attribute to image elements in model.

* `ImageStyleEngine`
Plugin that will add new converters converting style attribute from model to proper CSS classes in the view. It will also register `ImageStyleCommand` in the editor.

* `ImageToolbar`
Plugin implementing image toolbar: https://github.com/ckeditor/ckeditor5-ui-default/issues/136. It will allow to add items to it, will show/hide toolbar when needed.

* `ImageStyle`
Plugin that will connect the image toolbar with styles from configuration, it will add buttons that will execute proper command when clicked.

#### Configuration
Styles should be loaded from configuration. Configuration format, default values:
```js
config.define( 'image.styles', { 
    options: [ 
        { title: 'Align left', icon: 'image-align-left', value: 'left' }, 
        { title: 'Full', icon: 'image-align-full', value: false }, 
        { title: 'Align right', icon: 'image-align-right', value: 'right' } 
    ],
    className: value => `image-align-${ value }`
} );
```

Note: this configuration allows only to specify styles that exclude each other (there can be only one CSS class applied at given time). This can be solved in future by adding `optionGroups` configuration or specify a `key` for given group.

### LINKED ISSUE #5040: Image styles
During the talk with @Reinmar about image alignment, it turned out that we could create more general solution to apply any styles (CSS classes) to images, and not only alignment ones.
It should introduce new classes:

* `ImageStyleCommand`
Command that will apply proper style attribute to image elements in model.

* `ImageStyleEngine`
Plugin that will add new converters converting style attribute from model to proper CSS classes in the view. It will also register `ImageStyleCommand` in the editor.

* `ImageToolbar`
Plugin implementing image toolbar: https://github.com/ckeditor/ckeditor5-ui-default/issues/136. It will allow to add items to it, will show/hide toolbar when needed.

* `ImageStyle`
Plugin that will connect the image toolbar with styles from configuration, it will add buttons that will execute proper command when clicked.

#### Configuration
Styles should be loaded from configuration. Configuration format, default values:
```js
config.define( 'image.styles', { 
    options: [ 
        { title: 'Align left', icon: 'image-align-left', value: 'left' }, 
        { title: 'Full', icon: 'image-align-full', value: false }, 
        { title: 'Align right', icon: 'image-align-right', value: 'right' } 
    ],
    className: value => `image-align-${ value }`
} );
```

Note: this configuration allows only to specify styles that exclude each other (there can be only one CSS class applied at given time). This can be solved in future by adding `optionGroups` configuration or specify a `key` for given group.

szymonkups:
Fixes ckeditor/ckeditor5#5040.
~For now icons are added inside this PR, this will change when common icons will land in core: https://github.com/ckeditor/ckeditor5-theme-lark/issues/59.~

EDIT:
~For now core icons `align-center.svg` and `align-right.svg` are used. This will be changed to when proper icons land in core: https://github.com/ckeditor/ckeditor5-theme-lark/issues/59.~

EDIT v2: Proper icons are used.


src/imagestyle/imagestyle.js
```diff
@@ -0,0 +1,76 @@
...
+	 */
+	init() {
+		const styles = this.editor.config.get( 'image.styles' );
+
+		for ( let style of styles ) {
+			this._createButton( style );
+		}
+
+		// If there is no default image toolbar configuration, add all image styles buttons.
+		const imageToolbarConfig = this.editor.config.get( 'image.toolbar' );
```
Reinmar: This is not going to work this way... Because, if there's any other plugin which would like to extend this toolbar, it won't do that, because the toolbar is already set by this plugin.

So, unfortunately, we must make this a bit more complex.

1. The assumption is that `ImageToolbar` is loaded and sets `image.defaultToolbar` to an empty array in its constructor (which is "before init"). If it's not loaded, the property doesn't exist.
2. If the property exists, the other plugins should then **push** their default buttons to this array in their `init()` callbacks.
3. The ImageToolbar should then use `image.toolbar` or `image.defaultToolbar`.

Sounds reasonable?

src/imagestyle/imagestylecommand.js
```diff
@@ -0,0 +1,118 @@
...
+ */
+export default class ImageStyleCommand extends Command {
+	/**
+	 * Creates instance of the command.
+	 *
+	 * @param {module:core/editor/editor~Editor} editor Editor instance.
+	 * @param {Array.<module:image/imagestyle/imagestyleengine~ImageStyleFormat>} styles Allowed styles.
+	 */
+	constructor( editor, styles ) {
+		super( editor );
```
Reinmar, line 26: Missing new line.

src/imagestyle/imagestyle.js
```diff
@@ -0,0 +1,76 @@
...
+	_createButton( style ) {
+		const editor = this.editor;
+		const command = editor.commands.get( 'imagestyle' );
+		const t = editor.t;
+
+		editor.ui.componentFactory.add( style.name, ( locale ) => {
+			const view = new ButtonView( locale );
+
+			view.set( {
+				label: t( style.title ),
```
Reinmar: It can't be used like that. `t()` must be always used with a string. So the `t()` function should be used here: https://github.com/ckeditor/ckeditor5-image/pull/23/files#diff-00366adf927974c568c9ea53d58ede26R44

src/imagestyle/imagestyle.js
```diff
@@ -0,0 +1,76 @@
...
+				icon: style.icon
```
Reinmar: You can now also add `tooltip: true`

src/imagestyle/imagestyleengine.js
```diff
@@ -0,0 +1,90 @@
...
+		// Converter for figure element from view to model.
+		for ( let style of styles ) {
+			// Create converter only for non-null values.
+			if ( style.value !== null ) {
+				data.viewToModel.on( 'element:figure', viewToModelImageStyle( style ), { priority: 'low' } );
+			}
+		}
+
+		// Register image style command.
+		editor.commands.set( 'imagestyle', new ImageStyleCommand( editor, styles ) );
```
Reinmar: We're rather going to split the heading command into multiple heading commands, so the same should happen here. See https://github.com/ckeditor/ckeditor5-heading/issues/53. There should be a separate command per each style.

src/imagestyle/imagestyleengine.js
```diff
@@ -0,0 +1,90 @@
...
+	}
+}
+
+/**
+ * Image style format descriptor.
```
Reinmar, line 78: Some example would do nicely :P

src/imagestyle/utils.js
```diff
@@ -0,0 +1,36 @@
...
+import ModelElement from 'ckeditor5-engine/src/model/element';
+
+/**
+ * Checks if provided modelElement is an instance of {@link module:engine/model/element~Element Element} and its name
+ * equals to `image`.
+ *
+ * @param {module:engine/model/element~Element} modelElement
+ * @returns {Boolean}
+ */
+export function isImage( modelElement ) {
```
Reinmar: Would rather make for some base image utils. I can see there `src/utils.js`.

src/imagestyle/imagestyle.js
szymonkups: You're right. I didn't take other plugins into consideration. So we will use `image.defaultToolbar` configuration only if ` image.toolbar` is not specified. Is that correct?

src/imagestyle/utils.js
```diff
@@ -0,0 +1,36 @@
...
+}
+
+/**
+ * Returns style with given `value` from array of styles.
+ *
+ * @param {String} value
+ * @param {Array.<module:image/imagestyle/imagestyleengine~ImageStyleFormat> } styles
+ * @return {module:image/imagestyle/imagestyleengine~ImageStyleFormat|undefined}
+ */
+export function getStyleByValue( value, styles ) {
```
Reinmar: I guess this won't be necessary after the proposed change to split the command.

src/imagetoolbar.js
```diff
@@ -0,0 +1,127 @@
...
+		// Create a plain toolbar instance.
+		const toolbar = new ToolbarView();
+
+		// Create a BalloonPanelView instance.
+		const panel = new BalloonPanelView( editor.locale );
+
+		Template.extend( panel.template, {
+			attributes: {
+				class: [
+					'ck-toolbar__container',
```
Reinmar: Trailing comma.

Reinmar (changes requested):
The feature works great and there are just 2 rather simple code refactorings to do – the command split and how the default config is created.

src/imagestyle/imagestyle.js
Reinmar: Yes.

src/imagestyle/utils.js
szymonkups: It is also used by converters to be sure if applied value is correct with provided configuration.

szymonkups: But now it is used only there and looks like it can be just a private to that module. This way we can get rid of utils file.

src/imagestyle/converters.js
```diff
@@ -0,0 +1,98 @@
...
+
+		if ( eventType == 'addAttribute' || eventType == 'changeAttribute' ) {
+			if ( !newStyle ) {
+				return;
+			}
+
+			viewElement.addClass( newStyle.className );
+		}
+
+		consumable.consume( data.item, consumableType );
```
Reinmar: Why is it kept here? Can't it replace the `consumable.test()` call?

src/imagestyle/converters.js
```diff
@@ -0,0 +1,98 @@
...
+			return;
+		}
+
+		// Check if there is class name associated with given value.
+		const newStyle = getStyleByValue( data.attributeNewValue, styles );
+		const oldStyle = getStyleByValue( data.attributeOldValue, styles );
+		const viewElement = conversionApi.mapper.toViewElement( data.item );
+
+		if ( eventType == 'changeAttribute' || eventType == 'removeAttribute' ) {
+			if ( !oldStyle ) {
```
Reinmar: I don't get it... why do you check it here?

Reinmar: The style attribute is used by this feature. If the `getStyleByValue` method can't find it, then something's really broken and we're just hiding the bug this way.

src/imagestyle/imagestyleengine.js
```diff
@@ -0,0 +1,102 @@
...
+		editing.modelToView.on( 'changeAttribute:imageStyle:image', modelToViewConverter );
+		data.modelToView.on( 'changeAttribute:imageStyle:image', modelToViewConverter );
+		editing.modelToView.on( 'removeAttribute:imageStyle:image', modelToViewConverter );
+		data.modelToView.on( 'removeAttribute:imageStyle:image', modelToViewConverter );
+
+		for ( let style of styles ) {
+			// Converter for figure element from view to model.
+			// Create converter only for non-null values.
+			if ( style.value !== null ) {
+				data.viewToModel.on( 'element:figure', viewToModelImageStyle( style ), { priority: 'low' } );
```
Reinmar: Can't we have a single listener for all styles? Like for the opposite conversion?

src/imagestyle/imagestylecommand.js
```diff
@@ -0,0 +1,109 @@
...
+
+		const editor = this.editor;
+		const doc = editor.document;
+		const selection = doc.selection;
+		const imageElement = selection.getSelectedElement();
+
+		doc.enqueueChanges( () => {
+			const batch = options.batch || doc.batch();
+
+			batch.setAttribute( imageElement, 'imageStyle', this.style.value );
```
Reinmar, line 106: The attribute should be called `'style'`. It's image element's attribute after all.

Reinmar (changes requested):
Some questions regarding conversion and one issue (name of the attribute).

src/imagestyle/converters.js
szymonkups: I wasn't sure how to handle it to be honest. On the one hand we have `ImageStyle` plugin that have its configuration and I wanted to build a converter that will only handle values from that configuration. On the other hand - there won't be probably any other plugin that will write something to `imageStyle` attribute.

src/imagestyle/converters.js
szymonkups: I did some other checks and return points in this function so I wanted to test first and then consume when everything is validated.

src/imagestyle/imagestylecommand.js
szymonkups: I wanted to call it `'style'` but we have this: https://github.com/ckeditor/ckeditor5-engine/issues/559 that prevents us from using test utils.

src/imagestyle/converters.js
Reinmar: Actually, that condition could even be there, but if it wasn't so... distinguished :D. The code needs to be intuitive. Now, you read "if something, then ... if not something ABORT... do something". This isn't very natural of expressing this logic.

This would be better:

```js
if ( oldStyle && needToRemove( eventName ) ) {
   remove();
}
```

(See how you don't have to parse `eventName` in this callback and how well the code reads?)

And I agree – it's more error safe to check if `oldStyle` was found. But how was the value of the model element attribute set in the first place? There would need to be an additional feature which uses the same attribute – I don't think this should happen. But I may be wrong :P.

Anyway, since you have tests for this already, then we can keep it. Just make the code more intuitive. Also, you could consider splitting the function to two – one to handle change and remove attribute and other for add and change.

I also wrote a comment in https://github.com/ckeditor/ckeditor5-engine/issues/736 that we need better tools to write converters, because, e.g. in this case, there are too many listeners needed. Let's keep collecting use case to simplify.

src/imagestyle/imagestyleengine.js
szymonkups: That's a good point, I'll try to do this with one listener.

src/imagestyle/imagestylecommand.js
Reinmar: That's a very bad reason for choosing this name :P. Let's see if we can do something to fix that bug. It's a very serious problem.

Reinmar: Note: This is quite important what attribute name we use because this is kind of a public API of this feature.

szymonkups: > That's a very bad reason for choosing this name :P. Let's see if we can do something to fix that bug. It's a very serious problem.

I will fix other issues with this PR and then look into that issue.

szymonkups: I've created separate issue for that: https://github.com/ckeditor/ckeditor5-image/issues/25.

PR #46, opened 2017-02-01
Define more complete default image styles

### LINKED ISSUE #44: Improve the default image styles
First of all – the `` must be a block. Otherwise, it depends on `line-height` and `font-size` which causes ugly vertical margins to appear. This was prevented in ckeditor5.github.io by setting `font-size: 0` which then broke `margin-left: 1em`.

Other than that, image side style could define some max-width. And it should not define some odd `display` IMO.

### LINKED ISSUE #5062: Improve the default image styles
First of all – the `` must be a block. Otherwise, it depends on `line-height` and `font-size` which causes ugly vertical margins to appear. This was prevented in ckeditor5.github.io by setting `font-size: 0` which then broke `margin-left: 1em`.

Other than that, image side style could define some max-width. And it should not define some odd `display` IMO.

Reinmar:
Fixes ckeditor/ckeditor5#5062.

I was also able to remove this from ckeditor5.github.io:

```
-.ck-editor .ck-widget {
-    padding: 0;
-    line-height: 0;
-    font-size: 0;
-}
```

### DIFF
--- a/theme/theme.scss
+++ b/theme/theme.scss
@@ -20,13 +20,21 @@
 }
 
 // Image widget's styles.
-.ck-widget.image {
+.image {
 	text-align: center;
 	clear: both;
 
 	&.image-style-side {
-		display: inline-block;
 		float: right;
 		margin-left: 0.8em;
+		max-width: 50%;
 	}
+}
+
+.image > img {
+	// Prevent unnecessary margins caused by line-height (see #44).
+	display: block;
+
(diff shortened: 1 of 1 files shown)


Reinmar:
I added the `margin: 0 auto` style for the `img` to satisfy our manual tests where the image is super small and it wasn't centered without that. It's not crucial to have that style, but I guess it makes sense in general too.

PR #48, opened 2017-02-02
Image captioning

### LINKED ISSUE #28: Image captioning
First topic to discuss – what should be the data model and how we can implement the behaviour that the caption appears once you focus the image.

There are a couple of options, but none seems to be perfect:

1. Handle this upon conversion – in the model there will be no `<caption>` element when the image doesn't have caption. When you focus the image, the caption converter (hooked to selection change event) will add it to the *view*. This means, that the `<caption>` will be added to the model once you start typing and should be removed once you delete all the text (even before you blur the caption).
2. The second option is to always create the `<image>` element with a `<caption>` inside in the model, regardless of whether it has any content or not. And then, if it's empty, don't render it to the view.

The problems with the 1st option:

* You can have a `<figcaption>` element in the view, but no `<caption>` in the model, which means that there will be unmapped elements. This will be especially weird if you can place the selection in the rendered `<figcaption>`, but you can't map it to the model (cause there's no `<caption>` there).

The problems with the 2nd option:

* How can we force creating `<caption>` every time an `<image>` is created? These features don't know much about themselves and it needs to be automated somehow. E.g. the image upload feature will not know whether it should create the caption or not. The possible solution could be to create a post-fix listener on the model's insert event. The listener would add missing captions. It would be registered by the image caption feature.

  This may not sound like an elegant solution, but in fact, that's how we wanted to implement rules which a static schema will not able to enforce. There were two options in the past:

  * Either the schema would contain logic which is able to fix the model after e.g. OT broke it.
  * Or we can have a static schema (as a definition what is correct and what not) and listeners which post-fix broken model.

  So, the propose solution is in line with that.

### LINKED ISSUE #5048: Image captioning
First topic to discuss – what should be the data model and how we can implement the behaviour that the caption appears once you focus the image.

There are a couple of options, but none seems to be perfect:

1. Handle this upon conversion – in the model there will be no `<caption>` element when the image doesn't have caption. When you focus the image, the caption converter (hooked to selection change event) will add it to the *view*. This means, that the `<caption>` will be added to the model once you start typing and should be removed once you delete all the text (even before you blur the caption).
2. The second option is to always create the `<image>` element with a `<caption>` inside in the model, regardless of whether it has any content or not. And then, if it's empty, don't render it to the view.

The problems with the 1st option:

* You can have a `<figcaption>` element in the view, but no `<caption>` in the model, which means that there will be unmapped elements. This will be especially weird if you can place the selection in the rendered `<figcaption>`, but you can't map it to the model (cause there's no `<caption>` there).

The problems with the 2nd option:

* How can we force creating `<caption>` every time an `<image>` is created? These features don't know much about themselves and it needs to be automated somehow. E.g. the image upload feature will not know whether it should create the caption or not. The possible solution could be to create a post-fix listener on the model's insert event. The listener would add missing captions. It would be registered by the image caption feature.

  This may not sound like an elegant solution, but in fact, that's how we wanted to implement rules which a static schema will not able to enforce. There were two options in the past:

  * Either the schema would contain logic which is able to fix the model after e.g. OT broke it.
  * Or we can have a static schema (as a definition what is correct and what not) and listeners which post-fix broken model.

  So, the propose solution is in line with that.

szymonkups:
Adds captioning feature.
Currently there is no placeholder when caption's nested editable is empty.
There are other issues with nested editables, that will be reported in proper repositories and fixed (keyboard support etc.).
Fixes ckeditor/ckeditor5#5048.


Reinmar:
Some issues to be resolved in the first place:

1. There are too few features in the manual test. It's hard to test what happens in the caption. Tere are neither image features loaded, nor other features.
2. When the selection is inside the nested editable, the widget should not have the focus outline. There can be just one thing focus at a time and, in such situation, it's the nested editable.
3. See the screenshot – on this branch the image is rendered over the toolbar, when on master it's below:

  image

Reinmar:
And tests fail after I merged master into this branch.

src/converters.js
```diff
@@ -57,6 +59,13 @@ export function viewToModelImage() {
 			modelImage.setAttribute( 'alt', viewImg.getAttribute( 'alt' ) );
 		}
 
+		// Convert children of converted view element and append them to `modelImage`.
```
Reinmar, line 62: Please link to https://github.com/ckeditor/ckeditor5-engine/issues/736

src/imagecaptioning/imagecaptioning.js
```diff
@@ -0,0 +1,26 @@
...
+import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
+import ImageCaptioningEngine from './imagecaptioningengine';
+import '../../theme/imagecaptioning/theme.scss';
+
+/**
+ * The image captioning plugin.
+ *
+ * @extends module:core/plugin~Plugin
+ */
+export default class ImageCaptioning extends Plugin {
```
Reinmar: Simply: `ImageCaption`. We have `ImageStyles` not `ImageStyling`.

Reinmar: Besides, since this plugin is empty now, you can merge the engine part into it. Unless... unless we see that adding the placeholder will require some strictly UI-related logic.

OTOH, by splitting the feature beforehand we stick to some pattern, so I'm not sure what would be better.

src/imagecaptioning/imagecaptioningengine.js
```diff
@@ -0,0 +1,224 @@
...
+		 * @member {module:image/imagecaptioning/imagecaptioningengine~ImageCaptioningEngine} #_lastSelectedEditable
+		 */
+
+		// Schema configuration.
+		schema.registerItem( 'caption' );
+		schema.allow( { name: '$inline', inside: 'caption' } );
+		schema.allow( { name: 'caption', inside: 'image' } );
+
+		// Add caption element to each image inserted without it.
+		document.on( 'change', insertCaptionElement );
```
Reinmar: Maybe call this function `insertMissingCaptionElement`.

src/imagecaptioning/utils.js
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Returns function that creates caption editable element for given {@link module:engine/view/document~Document}.
+ *
+ * @param {module:engine/view/document~Document} viewDocument
+ * @return {Function}
+ */
+export function captionEditableCreator( viewDocument ) {
+	return () => {
+		const editable = new ViewEditableElement( 'figcaption', { contenteditable: true } ) ;
```
Reinmar: Space before `;`.

src/imagecaptioning/utils.js
```diff
@@ -0,0 +1,63 @@
...
+
+const captionSymbol = Symbol( 'imageCaption' );
+
+/**
+ * Returns function that creates caption editable element for given {@link module:engine/view/document~Document}.
+ *
+ * @param {module:engine/view/document~Document} viewDocument
+ * @return {Function}
+ */
+export function captionEditableCreator( viewDocument ) {
```
Reinmar: `editableCaptionCreator()`

src/imagecaptioning/utils.js
```diff
@@ -0,0 +1,63 @@
...
+ * @module image/imagecaptioning/utils
+ */
+
+import ViewEditableElement from '@ckeditor/ckeditor5-engine/src/view/editableelement';
+import ModelElement from '@ckeditor/ckeditor5-engine/src/model/element';
+
+const captionSymbol = Symbol( 'imageCaption' );
+
+/**
+ * Returns function that creates caption editable element for given {@link module:engine/view/document~Document}.
```
Reinmar: "Returns a function", "for the given"

src/imagecaptioning/utils.js
```diff
@@ -0,0 +1,63 @@
...
+	};
+}
+
+/**
+ * Returns `true` if given view element is image's caption editable.
+ *
+ * @param {module:engine/view/element~Element} viewElement
+ * @return {Boolean}
+ */
+export function isCaptionEditable( viewElement ) {
```
Reinmar: `isEditableCaption` to pair with the other method. I'm also thinking about `isCaption` and `captionElementCreator`. That "editable" part is confusing. Or maybe `captionEditableElementCreator`? I don't know... :D So maybe let's choose the shortest versions, WDYT?

src/imagecaptioning/utils.js
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/**
+ * Returns caption's model element from given image element. Returns `null` if no caption is found.
+ *
+ * @param {module:engine/model/element~Element} imageModelElement
+ * @return {module:engine/model/element~Element|null}
+ */
+export function getCaptionFromImage( imageModelElement ) {
+	for ( let node of imageModelElement.getChildren() ) {
```
Reinmar: Could be like:

```js
Array.from( imageModelElement.getChildren() ).find( node => node.is( 'caption' ) )
```

If we do https://github.com/ckeditor/ckeditor5-engine/issues/809.

src/widget/widget.js
```diff
@@ -52,13 +53,21 @@ export default class Widget extends Plugin {
...
+		if ( domEventData.target instanceof ViewEditableElement ) {
+			return;
+		}
+
 		// If target is not a widget element - check if one of the ancestors is.
+		let widgetElement = domEventData.target;
+
 		if ( !isWidget( widgetElement ) ) {
-			widgetElement = widgetElement.findAncestor( element => isWidget( element ) );
+			widgetElement = widgetElement.findAncestor( element => {
```
Reinmar: `widgetElement.findAncestor( isWidget );`

src/widget/widgetengine.js
```diff
@@ -34,6 +35,22 @@ export default class WidgetEngine extends Plugin {
...
 			const viewSelection = conversionApi.viewSelection;
+
+			// Add CSS class if selection is placed inside nested editable that belongs to widget.
+			const editableElement = viewSelection.editableElement;
+
+			if ( editableElement && !( editableElement instanceof RootEditable ) ) {
+				const widget = editableElement.findAncestor( element => isWidget( element ) );
+
+				if ( widget ) {
+					widget.addClass( WIDGET_SELECTED_CLASS_NAME );
```
Reinmar: Do we use the `I_AM_CONST` notation? I don't remember now what we've decided. cc @pjasiun, @scofalik

scofalik: This is complicated :P.
Instead of using stuff like:
```
const STICKS_TO_PREVIOUS = 0;
const STICKS_TO_NEXT = 1;
...
if ( this.stickiness == STICKS_TO_PREVIOUS ) { ... }
```
we decided to just use `'sticksToPrevious'` string. But this works only for closed list of constant values and was done to get rid of imports.

I don't know whether we decided anything about package-scope/configuration constants like this. If I wrote that code, I'd go with `widgetSelectedClassName` but I can't say this is wrong either.

Reinmar: I can see we use the ALL_CAPS notation in some cases. `INLINE_FILLER_LENGTH`, `BR_FILLER`, `NBSP_FILLER`, `DEFAULT_PRIORITY`. Also, all but one `export const` use this notation. So let's stick to that.

pjasiun:
Can't you, instead of `createImageAttributeConverter` and `modelToViewAttributeConverter` use model to view conversion builder with `fromAttribute( ... ).toAttribute( ... );`?

szymonkups:
>Can't you, instead of createImageAttributeConverter and modelToViewAttributeConverter use model to view conversion builder with fromAttribute( ... ).toAttribute( ... );?

Conversion builder does not allow to specify any information about attribute's element. I don't want to convert all `src` and `alt` attributes, only those from `image` elements.

src/imagecaptioning/utils.js
szymonkups: I'll go with `captionElementCreator` and `isCaption` - sounds better and it's shorter too.

szymonkups:
Added caption to limits set in schema, test it further after closing https://github.com/ckeditor/ckeditor5-engine/pull/825.

Reinmar:
Everything works brilliantly!

Reinmar:
I'm closing this PR despite the fact that the related tickets in engine and enter packages are still open. Let's not risk some conflicts.

I've spotted two issues with the image caption implementation: #58 and #57.

PR #49, opened 2017-02-03
Moved ImageBalloonPanelView focus tracking to init() method

### LINKED ISSUE #42: Fix the extend-template-render warning
UI lib introduced a warning about extending already rendered templates and it turns out that the image feature logs a lot of them:

https://travis-ci.org/ckeditor/ckeditor5-image/builds/195804693#L1452

### LINKED ISSUE #5060: Fix the extend-template-render warning
UI lib introduced a warning about extending already rendered templates and it turns out that the image feature logs a lot of them:

https://travis-ci.org/ckeditor/ckeditor5-image/builds/195804693#L1452

szymonkups:
### Proposed merge commit message

Fix: Moved focus tracking setup in `ImageBalloonPanelView` to `init()` method. Closes ckeditor/ckeditor5#5060.

### DIFF
--- a/src/ui/imageballoonpanelview.js
+++ b/src/ui/imageballoonpanelview.js
@@ -54,9 +54,6 @@ export default class ImageBalloonPanelView extends BalloonPanelView {
 		this.editor = editor;
 		const editingView = editor.editing.view;
 
-		// Let the focusTracker know about new focusable UI element.
-		editor.ui.focusTracker.add( this.element );
-
 		// Hide the balloon if editor had focus and now focus is lost.
 		this.listenTo( editor.ui.focusTracker, 'change:isFocused', ( evt, name, is, was ) => {
 			if ( was && !is ) {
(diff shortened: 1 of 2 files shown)


src/ui/imageballoonpanelview.js
```diff
@@ -86,6 +83,16 @@ export default class ImageBalloonPanelView extends BalloonPanelView {
 	}
 
 	/**
+	 * @inheritDoc
+	 */
+	init() {
+		super.init();
```
Reinmar: This may return a promise. Switch the order of calls in this function and return the result of `super.init()` – that'd be the easiest solution. And remember to write a test that a promise is returned ;)

Reinmar: PS. I wouldn't require that if `BalloonPanelView` was from this package. But it's a separate package so it can change independently.

PR #51, opened 2017-02-14
Enhancement to fake selection label handing in widgets

### LINKED ISSUE #9: Fake selection label handling
Quoting @Reinmar's PR comment:
> There should be no special selection converter for image widget. It should be general to widgets. So that label must be handled differently

Right now adding a fake selection label is done by special selection converter. This should be simplified, as similar converter is used on general widget implementation.

### LINKED ISSUE #5034: Fake selection label handling
Quoting @Reinmar's PR comment:
> There should be no special selection converter for image widget. It should be general to widgets. So that label must be handled differently

Right now adding a fake selection label is done by special selection converter. This should be simplified, as similar converter is used on general widget implementation.

szymonkups:

Fix: enhancement to fake selection label handing in widgets. Closes ckeditor/ckeditor5#5034.

---

### Additional information

Removed selection conversion from image itself, all handling is done inside widget feature. Widget utils expose two new methods: `setFakeSelectionLabel` and `setFakeSelectionLabel` which can be used to attach fake selection label to element and retrieve it. Widget selection converter uses it to get proper fake selection label, so image feature (and other widget features in feature) need only care about `setFakeSelectionLabel`.


src/widget/utils.js
```diff
@@ -52,6 +53,34 @@ export function widgetize( element ) {
...
+ * Returns fake selection label for provided element.
+ *
+ * @param {module:engine/view/element~Element} element
+ * @return {String|undefined}
+ */
+export function getFakeSelectionLabel( element ) {
+	const labelCreator = element.getCustomProperty( fakeSelectionLabelSymbol );
+
+	if ( !labelCreator ) {
+		return undefined;
```
Reinmar: An empty string would be better I think. Let's say that there's always some label but sometimes it's empty :).

src/utils.js
```diff
@@ -7,22 +7,36 @@
...
  * @param {module:engine/view/element~Element} viewElement
+ * @param {function} t Shorthand for {@link module:utils/locale~Locale#t}.
  * @returns {module:engine/view/element~Element}
  */
-export function toImageWidget( viewElement ) {
+export function toImageWidget( viewElement, t ) {
 	viewElement.setCustomProperty( imageSymbol, true );
 
+	setFakeSelectionLabel( viewElement, () => {
+		let fakeSelectionLabel = t( 'image widget' );
```
Reinmar: Why isn't the label passed instead of the `t()` function? This is something which the feature should care about, not this util function.

src/widget/utils.js
```diff
@@ -52,6 +53,34 @@ export function widgetize( element ) {
...
 
+/**
+ * Sets fake selection label for given element.
+ * It can be passed as a plain string or a function returning a string. Function will be called each time label is retrieved by
+ * {module:image/widget/utils~getFakeSelectionLabel}.
+ *
+ * @param {module:engine/view/element~Element} element
+ * @param {String|Function} labelOrCreator
+ */
+export function setFakeSelectionLabel( element, labelOrCreator ) {
```
Reinmar: I think that we can keep these functions, but it'd be easier if the label could be passed straight to `widgetize()`, in one of the options.

Besides, let's use a more generic function name – `setLabel()`. How the label is used is a different story and shouldn't be a developer's concern.

Reinmar:
R-, but I've noticed only a minor issues regarding where a specific piece of code would make more sense.

src/image/imageengine.js
```diff
@@ -54,16 +54,13 @@ export default class ImageEngine extends Plugin {
 		// Build converter from model to view for editing pipeline.
 		buildModelConverter().for( editing.modelToView )
 			.fromElement( 'image' )
-			.toElement( () => toImageWidget( createImageViewElement() ) );
+			.toElement( () => toImageWidget( createImageViewElement(), editor.t( 'image widget' ) ) );
```
Reinmar: The `t()` function should be used directly (`editor.t()` => `t()`).

Reinmar:
I'll fix the remaining issue and close the PR.

Reinmar:
Unfortunately, there's a bug. The label is changing every time I focus a widget:

image

Reinmar:
Hm... but it can be a bug in the engine.

Reinmar:
Yep, when rendering the selection we do this:

```
		this._fakeSelectionContainer.firstChild.data = content;
```

which is a bit unsafe if the text node got split for some reason. I don't know if it happens, but you can test this hyphothesis.

Reinmar:
I can't stop testing it :D

> which is a bit unsafe if the text node got split for some reason. I don't know if it happens, but you can test this hyphothesis.

I wasn't right. The `selection.fakeSelectionLabel` contains the duplicated string.

PR #70, opened 2017-03-13
Fix: The image goes beyond the boundaries of the parent block. Closes…

### LINKED ISSUE #67: Side image is not styled correctly
Side image is not styled correctly.

Go to http://localhost:8125/ckeditor5-presets/tests/manual/article.html and you'll see that the image overflow the widget.

image

### LINKED ISSUE #5073: Side image is not styled correctly
Side image is not styled correctly.

Go to http://localhost:8125/ckeditor5-presets/tests/manual/article.html and you'll see that the image overflow the widget.

image

oleq:
… ckeditor/ckeditor5#5073.

Fix: The image goes beyond the boundaries of the parent block. Closes ckeditor/ckeditor5#5073.

### DIFF
--- a/theme/theme.scss
+++ b/theme/theme.scss
@@ -37,4 +37,8 @@
 
 	// Center the image if its width is smaller than content's width.
 	margin: 0 auto;
+
+	// Make sure the image never exceeds the size of the parent container (#67).
+	max-width: 100%;
+	max-height: 100%;
 }
--- a/theme/theme.scss
+++ b/theme/theme.scss
@@ -40,5 +40,4 @@
 
 	// Make sure the image never exceeds the size of the parent container (#67).
 	max-width: 100%;
-	max-height: 100%;
 }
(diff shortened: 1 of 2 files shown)


theme/theme.scss
```diff
@@ -37,4 +37,8 @@
 
 	// Center the image if its width is smaller than content's width.
 	margin: 0 auto;
+
+	// Make sure the image never exceeds the size of the parent container (#67).
+	max-width: 100%;
+	max-height: 100%;
```
pjasiun: Are you sure we need `max-height`? For me, there is nothing wrong if the long image is longer then editors editable area. Also, it means that a long image will be resized if editable have dynamic height, isn't it?

oleq: Given the editor's height is not restricted by CSS (parent container), then yes. I wouldn't be so sure there is no such case, though. OTOH, you're probably right.

PR #74, opened 2017-03-17
Extracted nested editable styles and logic to widgets.

### LINKED ISSUE #57: Nested editables must have their classes
In https://github.com/ckeditor/ckeditor5-image/pull/48 we introduced first nested editables, but they are styled as:

```
 +// Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
 +// For licensing, see LICENSE.md or http://ckeditor.com/license
 +
 +@import '~@ckeditor/ckeditor5-theme-lark/theme/helpers/_colors.scss';
 +@import '~@ckeditor/ckeditor5-theme-lark/theme/helpers/_shadow.scss';
 +@import '~@ckeditor/ckeditor5-theme-lark/theme/helpers/_states.scss';
 +
 +.ck-widget.image {
 +	figcaption {
 +		background-color: ck-color( 'foreground' );
 +		padding: 10px;
 +
 +		// The `:focus` styles is applied before `.focused` class inside editables.
 +		// These styles show different border for a blink of an eye.
 +		&:focus {
 +			outline: none;
 +			box-shadow: none;
 +		}
 +
 +		&.focused {
 +			@include ck-focus-ring( 'outline' );
 +			@include ck-box-shadow( $ck-inner-shadow );
 +			background-color: ck-color( 'background' );;
 +		}
 +
 +	}
 +}
```

This is incorrect – part of these styles should be image styles and part of them should be generic widget styles. In order to style widget's nested editables in a generic way, we need to be able to identify nested editables in the DOM, hence, they need to be marked with classes.

### LINKED ISSUE #5067: Nested editables must have their classes
In https://github.com/ckeditor/ckeditor5-image/pull/48 we introduced first nested editables, but they are styled as:

```
 +// Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
 +// For licensing, see LICENSE.md or http://ckeditor.com/license
 +
 +@import '~@ckeditor/ckeditor5-theme-lark/theme/helpers/_colors.scss';
 +@import '~@ckeditor/ckeditor5-theme-lark/theme/helpers/_shadow.scss';
 +@import '~@ckeditor/ckeditor5-theme-lark/theme/helpers/_states.scss';
 +
 +.ck-widget.image {
 +	figcaption {
 +		background-color: ck-color( 'foreground' );
 +		padding: 10px;
 +
 +		// The `:focus` styles is applied before `.focused` class inside editables.
 +		// These styles show different border for a blink of an eye.
 +		&:focus {
 +			outline: none;
 +			box-shadow: none;
 +		}
 +
 +		&.focused {
 +			@include ck-focus-ring( 'outline' );
 +			@include ck-box-shadow( $ck-inner-shadow );
 +			background-color: ck-color( 'background' );;
 +		}
 +
 +	}
 +}
```

This is incorrect – part of these styles should be image styles and part of them should be generic widget styles. In order to style widget's nested editables in a generic way, we need to be able to identify nested editables in the DOM, hence, they need to be marked with classes.

szymonkups:

Fix: Nested editables have own CSS classes and their creation utility is extracted to widgets. Closes ckeditor/ckeditor5#5067.

BREAKING CHANGE: Widget utility function `widgetize` is renamed to `toWidget`.

### DIFF
--- a/src/widget/utils.js
+++ b/src/widget/utils.js
@@ -44,7 +44,7 @@ export function isWidget( element ) {
 }
 
 /**
- * "Widgetizes" given {@link module:engine/view/element~Element}:
+ * Converts given {@link module:engine/view/element~Element} to widget in following way:
  * * sets `contenteditable` attribute to `true`,
  * * adds custom `getFillerOffset` method returning `null`,
  * * adds `ck-widget` CSS class,
@@ -56,7 +56,7 @@ export function isWidget( element ) {
  * a plain string or a function returning a string.
  * @returns {module:engine/view/element~Element} Returns same element.
  */
-export function widgetize( element, options ) {
+export function toWidget( element, options ) {
 	options = options || {};
 	element.setAttribute( 'contenteditable', false );
 	element.getFillerOffset = getFillerOffset;
@@ -105,7 +105,7 @@ export function getLabel( element ) {
  * @param {module:engine/view/document~Document} viewDocument
  * @returns {module:engine/view/editableelement~EditableElement}
  */
-export function createNestedEditable( elementName, viewDocument ) {
+export function toWidgetEditable( elementName, viewDocument ) {
 	const editable = new ViewEditableElement( elementName, { contenteditable: true } );
 	editable.addClass( NESTED_EDITABLE_CLASS_NAME );
 	editable.document = viewDocument;
--- a/src/widget/utils.js
+++ b/src/widget/utils.js
@@ -20,13 +20,6 @@ const labelSymbol = Symbol( 'label' );
 export const WIDGET_CLASS_NAME = 'ck-widget';
 
 /**
- * CSS class added to each nested edtiable.
- *
- * @type {String}
- */
-export const NESTED_EDITABLE_CLASS_NAME = 'ck-nested-editable';
-
-/**
  * CSS class added to currently selected widget element.
  *
  * @const {String}
@@ -107,14 +100,14 @@ export function getLabel( element ) {
  */
 export function toWidgetEditable( elementName, viewDocument ) {
 	const editable = new ViewEditableElement( elementName, { contenteditable: true } );
-	editable.addClass( NESTED_EDITABLE_CLASS_NAME );
+	editable.addClass( 'ck-editable' );
 	editable.document = viewDocument;
 
 	editable.on( 'change:isFocused', ( evt, property, is ) => {
 		if ( is ) {
-			editable.addClass( 'focused' );
+			editable.addClass( 'ck-editable_focused' );
 		} else {
-			editable.removeClass( 'focused' );
+			editable.removeClass( 'ck-editable_focused' );
 		}
 	} );
 
--- a/src/widget/utils.js
+++ b/src/widget/utils.js
@@ -7,8 +7,6 @@
  * @module image/widget/utils
  */
 
-import ViewEditableElement from '@ckeditor/ckeditor5-engine/src/view/editableelement';
-
 const widgetSymbol = Symbol( 'isWidget' );
 const labelSymbol = Symbol( 'label' );
 
@@ -92,16 +90,17 @@ export function getLabel( element ) {
 }
 
 /**
- * Creates nested editable element with proper CSS classes.
+ * Adds functionality to provided {module:engine/view/editableelement~EditableElement} to act as a widget's editable:
+ * * sets `contenteditable` attribute to `true`,
+ * * adds `ck-editable` CSS class,
+ * * adds `ck-editable_focused` CSS class when editable is focused and removes it when it's blurred.
  *
- * @param {String} elementName Name of the element to be created.
- * @param {module:engine/view/document~Document} viewDocument
- * @returns {module:engine/view/editableelement~EditableElement}
+ * @param {module:engine/view/editableelement~EditableElement} editable
+ * @returns {module:engine/view/editableelement~EditableElement} Returns same element that was provided in `editable` param.
  */
-export function toWidgetEditable( elementName, viewDocument ) {
-	const editable = new ViewEditableElement( elementName, { contenteditable: true } );
+export function toWidgetEditable( editable ) {
+	editable.setAttribute( 'contenteditable', 'true' );
 	editable.addClass( 'ck-editable' );
-	editable.document = viewDocument;
 
 	editable.on( 'change:isFocused', ( evt, property, is ) => {
 		if ( is ) {
--- a/src/widget/utils.js
+++ b/src/widget/utils.js
@@ -7,6 +7,8 @@
  * @module image/widget/utils
  */
 
+import ViewEditableElement from '@ckeditor/ckeditor5-engine/src/view/editableelement';
+
 const widgetSymbol = Symbol( 'isWidget' );
 const labelSymbol = Symbol( 'label' );
 
@@ -16,6 +18,13 @@ const labelSymbol = Symbol( 'label' );
  * @const {String}
  */
 export const WIDGET_CLASS_NAME = 'ck-widget';
+
+/**
+ * CSS class added to each nested edtiable.
+ *
+ * @type {String}
+ */
+export const NESTED_EDITABLE_CLASS_NAME = 'ck-nested-editable';
 
 /**
  * CSS class added to currently selected widget element.
@@ -87,6 +96,29 @@ export function getLabel( element ) {
 	}
 
 	return typeof labelCreator == 'function' ? labelCreator() : labelCreator;
+}
+
+/**
+ * Creates nested editable element with proper CSS classes.
+ *
+ * @param {String} elementName Name of the element to be created.
+ * @param {module:engine/view/document~Document} viewDocument
+ * @returns {module:engine/view/editableelement~EditableElement}
+ */
+export function createNestedEditable( elementName, viewDocument ) {
+	const editable = new ViewEditableElement( elementName, { contenteditable: true } );
+	editable.addClass( NESTED_EDITABLE_CLASS_NAME );
+	editable.document = viewDocument;
+
+	editable.on( 'change:isFocused', ( evt, property, is ) => {
+		if ( is ) {
+			editable.addClass( 'focused' );
+		} else {
+			editable.removeClass( 'focused' );
+		}
+	} );
+
+	return editable;
 }
 
 // Default filler offset function applied to all widget elements.
--- a/theme/imagecaption/theme.scss
+++ b/theme/imagecaption/theme.scss
@@ -2,27 +2,12 @@
 // For licensing, see LICENSE.md or http://ckeditor.com/license
 
 @import '~@ckeditor/ckeditor5-theme-lark/theme/helpers/_colors.scss';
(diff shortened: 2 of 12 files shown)


src/widget/utils.js
```diff
@@ -89,6 +98,29 @@ export function getLabel( element ) {
...
 }
 
+/**
+ * Creates nested editable element with proper CSS classes.
+ *
+ * @param {String} elementName Name of the element to be created.
+ * @param {module:engine/view/document~Document} viewDocument
+ * @returns {module:engine/view/editableelement~EditableElement}
+ */
+export function createNestedEditable( elementName, viewDocument ) {
```
Reinmar: It would be more consistent with `widgetize()` if it accepted an element.

BTW, it'd be hard to follow "widgetize"'s namig style in this case, so perhaps we could use these names: `toWidget()` and `toWidgetEditable()`.

src/widget/utils.js
```diff
@@ -89,6 +98,29 @@ export function getLabel( element ) {
...
+	const editable = new ViewEditableElement( elementName, { contenteditable: true } );
+	editable.addClass( NESTED_EDITABLE_CLASS_NAME );
+	editable.document = viewDocument;
+
+	editable.on( 'change:isFocused', ( evt, property, is ) => {
+		if ( is ) {
+			editable.addClass( 'focused' );
```
Reinmar: Should be: `ck-editable_focused`.

See https://github.com/ckeditor/ckeditor5-design/wiki/Code-Style-Naming-Guidelines#css-classes

src/widget/utils.js
```diff
@@ -89,6 +89,30 @@ export function getLabel( element ) {
...
+ * Adds functionality to provided {module:engine/view/editableelement~EditableElement} to act as a widget's editable:
+ * * sets `contenteditable` attribute to `true`,
+ * * adds `ck-editable` CSS class,
+ * * adds `ck-editable_focused` CSS class when editable is focused and removes it when it's blurred.
+ *
+ * @param {module:engine/view/editableelement~EditableElement} editable
+ * @returns {module:engine/view/editableelement~EditableElement} Returns same element that was provided in `editable` param.
+ */
+export function toWidgetEditable( editable ) {
+	editable.setAttribute( 'contenteditable', 'true' );
```
oleq, line 102: I think this and the following line should be done via `Template.apply()`. One of the reasons we decided to have a `Template` class was that DOM attrs should be touched via unified API.

src/widget/utils.js
```diff
@@ -89,6 +89,30 @@ export function getLabel( element ) {
...
+	editable.addClass( 'ck-editable' );
+
+	editable.on( 'change:isFocused', ( evt, property, is ) => {
```
oleq, line 105:
```js
new Template( {
    attributes: {
        contenteditable: true,
        class: [ 
                'ck-editable', 
                bind.if( editable, 'isFocused', 'ck-editable_focused' ) 
        ]
    }
} ).apply( editable );
```

theme/imagecaption/theme.scss
```diff
@@ -2,27 +2,12 @@
 // For licensing, see LICENSE.md or http://ckeditor.com/license
 
 @import '~@ckeditor/ckeditor5-theme-lark/theme/helpers/_colors.scss';
-@import '~@ckeditor/ckeditor5-theme-lark/theme/helpers/_shadow.scss';
-@import '~@ckeditor/ckeditor5-theme-lark/theme/helpers/_states.scss';
 
 .ck-widget.image {
-	figcaption {
+	figcaption.ck-editable {
```
oleq: Does `figcaption` selector is really needed? Wouldn't `.ck-editable` be specific enough?

theme/widget/theme.scss
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
+// For licensing, see LICENSE.md or http://ckeditor.com/license
+
+@import '~@ckeditor/ckeditor5-theme-lark/theme/helpers/_colors.scss';
+@import '~@ckeditor/ckeditor5-theme-lark/theme/helpers/_shadow.scss';
+@import '~@ckeditor/ckeditor5-theme-lark/theme/helpers/_states.scss';
+
+.ck-widget {
+	margin: 10px 0;
```
oleq: I think that https://github.com/ckeditor/ckeditor5-theme-lark/blob/master/theme/helpers/_spacing.scss#L32-L34 would make more sense.

theme/widget/theme.scss
```diff
@@ -0,0 +1,37 @@
...
+	padding: 0;
+
+	&.ck-widget_selected, &.ck-widget_selected:hover {
+		outline: 2px solid #ace;
```
oleq: These colors `#ace`, `#ddd` and `yellow` should be variables. Or better, they should come from https://github.com/ckeditor/ckeditor5-theme-lark/blob/master/theme/helpers/_colors.scss#L21-L27. Or even better, the widgets could [extend the standard palette](https://github.com/ckeditor/ckeditor5-theme-lark/blob/master/theme/helpers/_colors.scss#L64-L66). We've never used it before, but it could still work:
```scss
@include ck-color-add( ( 
	'widget-selected': #ace, // 
	'widget-blurred': #ddd 
) );
```

then 

```scss
outline: 2px solid ck-color( 'widget-selected' );
```
BTW: We have [`focus` color](https://github.com/ckeditor/ckeditor5-theme-lark/blob/master/theme/helpers/_colors.scss#L23), which I think makes more sense than `#ace`.

theme/widget/theme.scss
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	.ck-editor__editable.ck-blurred &.ck-widget_selected {
+		outline: 2px solid #ddd;
```
oleq: `2px` is a magic number. This should be a variable on the top of the file. Also: see above.

theme/widget/theme.scss
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	&:hover {
+		outline: 2px solid yellow;
```
oleq: `2px` is a magic number. This should be a variable on the top of the file. Also: see above.

theme/widget/theme.scss
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	.ck-editable {
+		// The `:focus` styles is applied before `.focused` class inside editables.
```
oleq: Outdated docs (`.focused`?).

Also "styles is" vs. "styles are".

theme/widget/theme.scss
```diff
@@ -0,0 +1,37 @@
...
+		// These styles show different border for a blink of an eye.
+		&:focus {
```
oleq: Despite the comment, I don't quite get why we need this structure `:focus ..._focused`. A few more words would explain it better. Or do we need it at all?

theme/imagecaption/theme.scss
```diff
@@ -2,27 +2,12 @@
...
 		background-color: ck-color( 'foreground' );
 		padding: 10px;
```
oleq: `ck-spacing()`

theme/imagecaption/theme.scss
```diff
@@ -2,27 +2,12 @@
...
 		font-size: .8em;
```
oleq: `ck-font-size( -1 )` I guess.

oleq (changes requested):
Look fine but requires some improvements in styles.

src/widget/utils.js
Reinmar: That's not true. This isn't UI. This is the engine and the element here isn't a native DOM element. We've been talking about unifying the APIs, but that never happened and I guess won't happen anytime soon.

src/widget/utils.js
Reinmar: As above – nope. This is not a native element.

src/widget/utils.js
oleq: Nice catch! I didn't notice that this isn't DOM. You're right, ofc ;-)

theme/widget/theme.scss
szymonkups: This can be simplified to:
```CSS
.ck-editable {
	&.ck-editable_focused, &:focus {
		@include ck-focus-ring( 'outline' );
		@include ck-box-shadow( $ck-inner-shadow );
		background-color: ck-color( 'background' );
	}
}
```

theme/imagecaption/theme.scss
```diff
@@ -2,27 +2,14 @@
 // For licensing, see LICENSE.md or http://ckeditor.com/license
 
 @import '~@ckeditor/ckeditor5-theme-lark/theme/helpers/_colors.scss';
-@import '~@ckeditor/ckeditor5-theme-lark/theme/helpers/_shadow.scss';
-@import '~@ckeditor/ckeditor5-theme-lark/theme/helpers/_states.scss';
+@import '~@ckeditor/ckeditor5-theme-lark/theme/helpers/_spacing.scss';
+@import '~@ckeditor/ckeditor5-theme-lark/theme/helpers/_fonts.scss';
 
 .ck-widget.image {
```
Reinmar: It'd be better if these styles had nothing to do with the widget classes. The widgets theme use widget classes (and hence applies to the editor only) and the image theme uses image classes.

Actually, this is a bigger topic. For the widgets we're defining styles which will apply in a context of the editor cause we use editor classes like `.ck-widget`. For the content (like `figure.image`) we want to provide generic content styles like in the CKEditor 4's [`contents.css`](https://github.com/ckeditor/ckeditor-dev/blob/master/contents.css). So, they need to be overrideable but still so specific to apply to editor contents only, not the whole page.

The only idea I've got now is to prefix all of them with `.ck-editor__editable`. WDYT, @oleq?

oleq: Isn't `.ck-editor .image` enough? Because I think it is and it's a better DX because devs don't need to think what is "editable" and why. All they want is a styled **image** in the **CKEditor**.

Reinmar: Is `.ck-editor` applied in the inline editor? It'd be best if the class was as short as possible, but I think it can't be `.ck-editor`.

If we would like to use something shorter than `.ck-editor__editable`, then perhaps we can start using an additional class `.ck-content` for the editables?

oleq: > Is .ck-editor applied in the inline editor?

No. A good point. Let's proceed with `.ck-editor__editable` then.

Reinmar: Reported it for the future: https://github.com/ckeditor/ckeditor5-ui/issues/176.

theme/widget/theme.scss
```diff
@@ -0,0 +1,41 @@
...
+		outline: $widget-outline-thickness solid ck-color( 'focus' );
+	}
+
+	.ck-editor__editable.ck-blurred &.ck-widget_selected {
+		outline: $widget-outline-thickness solid ck-color( 'widget-blurred' );
+	}
+
+	&:hover {
+		outline: $widget-outline-thickness solid ck-color( 'widget-hover' );
+	}
```
Reinmar, line 30: Something's wrong here... Indentation! :P

szymonkups: :sob:

PR #81, opened 2017-03-23
T/77 Image captions in the view are hidden instead of being removed

### LINKED ISSUE #77: Yet another caption+undo bug: mapping empty caption.
**Steps to reproduce:**

1. Create image (must have empty caption).
2. Undo.

**Expected:** Image is removed.

**Actual:** Editor throws an error.

**Reason:**

To put things short and simple:

When model `<caption>` is empty, we remove `<figcaption>` element from view. But this causes a bug with undo:

1. When creating image, insert operation for caption is performed.
2. When undoing it, insert operation becomes remove operation for caption.
3. Remove operation has a range to remove, which is in `<figure>`.
4. However `<figcaption>` is not there, which throws an error (we are looking for a too big offset in `<figure>`).

It's easy to see that this is the problem, because you can undo the image insertion if you focus it.

**Solutions:**

There are two things we can do:

1. Introduce `remove:caption` converter which will check if the `<figcaption>` is there and consume change + stop event if it's not there.
2. Stop trying to fix all the issues that comes from the fact that we have element in model that sometimes is not in the view, and use CSS classes to hide it.

I am for 2.

### LINKED ISSUE #5080: Yet another caption+undo bug: mapping empty caption.
**Steps to reproduce:**

1. Create image (must have empty caption).
2. Undo.

**Expected:** Image is removed.

**Actual:** Editor throws an error.

**Reason:**

To put things short and simple:

When model `<caption>` is empty, we remove `<figcaption>` element from view. But this causes a bug with undo:

1. When creating image, insert operation for caption is performed.
2. When undoing it, insert operation becomes remove operation for caption.
3. Remove operation has a range to remove, which is in `<figure>`.
4. However `<figcaption>` is not there, which throws an error (we are looking for a too big offset in `<figure>`).

It's easy to see that this is the problem, because you can undo the image insertion if you focus it.

**Solutions:**

There are two things we can do:

1. Introduce `remove:caption` converter which will check if the `<figcaption>` is there and consume change + stop event if it's not there.
2. Stop trying to fix all the issues that comes from the fact that we have element in model that sometimes is not in the view, and use CSS classes to hide it.

I am for 2.

szymonkups:

Fix: Image captions in the view are hidden instead of being removed. Closes ckeditor/ckeditor5#5080 .

---

### Additional information

Please close #74 before.

### DIFF
--- a/src/imagecaption/imagecaptionengine.js
+++ b/src/imagecaption/imagecaptionengine.js
@@ -98,7 +98,7 @@ export default class ImageCaptionEngine extends Plugin {
 
 		// Hide last selected caption if have no child elements.
 		if ( this._lastSelectedCaption && !this._lastSelectedCaption.childCount ) {
-			this._lastSelectedCaption.addClass( 'ck-editable_hidden' );
+			this._lastSelectedCaption.addClass( 'ck-hidden' );
 		}
 
 		// If whole image widget is selected.
@@ -114,7 +114,7 @@ export default class ImageCaptionEngine extends Plugin {
 		}
 
 		if ( viewCaption ) {
-			viewCaption.removeClass( 'ck-editable_hidden' );
+			viewCaption.removeClass( 'ck-hidden' );
 			this._lastSelectedCaption = viewCaption;
 		}
 	}
@@ -171,7 +171,7 @@ function captionModelToView( elementCreator, hide = true ) {
 
 			// Hide if empty.
 			if ( !captionElement.childCount ) {
-				viewCaption.addClass( 'ck-editable_hidden' );
+				viewCaption.addClass( 'ck-hidden' );
 			}
 
 			insertViewCaptionAndBind( viewCaption, data.item, viewImage, conversionApi.mapper );
--- a/src/imagecaption/imagecaptionengine.js
+++ b/src/imagecaption/imagecaptionengine.js
@@ -42,7 +42,6 @@ export default class ImageCaptionEngine extends Plugin {
 		const schema = document.schema;
 		const data = editor.data;
 		const editing = editor.editing;
-
 		/**
 		 * Last selected caption editable.
 		 * It is used for hiding editable when is empty and image widget is no longer selected.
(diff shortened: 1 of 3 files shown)


src/imagecaption/imagecaptionengine.js
```diff
@@ -77,102 +75,47 @@ export default class ImageCaptionEngine extends Plugin {
...
 
-		if ( selectedElement && isImageWidget( selectedElement ) ) {
-			const viewImage = viewCaptionElement.findAncestor( element => element == selectedElement );
-
-			if ( viewImage ) {
-				return;
-			}
+		// Hide last selected caption if have no child elements.
+		if ( this._lastSelectedCaption && !this._lastSelectedCaption.childCount ) {
+			this._lastSelectedCaption.addClass( 'ck-editable_hidden' );
```
oleq: We have `.ck-hidden` class for this kind of things.

src/imagecaption/imagecaptionengine.js
```diff
@@ -246,20 +169,16 @@ function captionModelToView( elementCreator ) {
 				elementCreator.clone( true ) :
 				elementCreator();
 
+			// Hide if empty.
+			if ( !captionElement.childCount ) {
+				viewCaption.addClass( 'ck-editable_hidden' );
```
oleq: We have `.ck-hidden` class for this kind of things.

Reinmar:
I asked @scofalik to check this PR.

src/imagecaption/imagecaptionengine.js
```diff
@@ -201,42 +144,22 @@ function insertMissingModelCaptionElement( evt, changeType, data, batch ) {
...
 // @return {Function}
-function captionModelToView( elementCreator ) {
+function captionModelToView( elementCreator, hide = true ) {
 	return ( evt, data, consumable, conversionApi ) => {
 		const captionElement = data.item;
 
-		if ( isImage( captionElement.parent ) && ( captionElement.childCount > 0 ) ) {
+		// Return if element shouldn't be present when empty.
+		if ( !captionElement.childCount && !hide ) {
+			return;
```
scofalik, line 190: I think it would be better if this part of code is in `if ( isImage( ... ) ) { ... }` so it won't ever affect other converters if we have different type of captions.

scofalik: On the other hand it won't affect other converters because it does not consume anything, but I was also thinking whether the caption should not be consumed in this case 🤔 . Since we are not consuming, the code could stay here.

scofalik:
I'd like to see one improvement introduced here (unless there is already ticket for this and you plan to work on it). Right now, when content is inserted into the caption, the caption does not show. The content could come from OT, for example.

PR #83, opened 2017-03-29
T/78 Prevent adding caption to image twice

### LINKED ISSUE #78: Caption is added twice.
At the moment, there is a callback that adds `<caption>` to `<image>` whenever `<image>` without `<caption>` is inserted to the document. Unfortunately, there is a small bug in that fixer, which causes adding redundant `<caption>` (and it also makes the editor throw :P).

So the problem is when the change comes from OT (redo is taken care of because fixer checks `type == 'insert'`). Client A adds an image and generates `<caption>` using fixer.

Then on client B, first, you get the `InsertOperation` that inserts just the `<image>`. The fixer checks it and sees that empty `<image>` is added, so it adds the `<caption>`. But then you get the "fixer operation" from client A which adds additional `<caption>`.

There are two solutions possible.

1. In `batch.document.enqueueChanges` block, check if the `<image>` is still empty, before adding second `<caption>`.
2. Do not fix if `batch.type` is `transparent`.

I prefer 1 solution, at least for now. We haven't discussed how fixers and problems like this should be solved for OT yet (similar case will be with possible auto embed feature -- converted link should not be replaced twice).

### LINKED ISSUE #5082: Caption is added twice.
At the moment, there is a callback that adds `<caption>` to `<image>` whenever `<image>` without `<caption>` is inserted to the document. Unfortunately, there is a small bug in that fixer, which causes adding redundant `<caption>` (and it also makes the editor throw :P).

So the problem is when the change comes from OT (redo is taken care of because fixer checks `type == 'insert'`). Client A adds an image and generates `<caption>` using fixer.

Then on client B, first, you get the `InsertOperation` that inserts just the `<image>`. The fixer checks it and sees that empty `<image>` is added, so it adds the `<caption>`. But then you get the "fixer operation" from client A which adds additional `<caption>`.

There are two solutions possible.

1. In `batch.document.enqueueChanges` block, check if the `<image>` is still empty, before adding second `<caption>`.
2. Do not fix if `batch.type` is `transparent`.

I prefer 1 solution, at least for now. We haven't discussed how fixers and problems like this should be solved for OT yet (similar case will be with possible auto embed feature -- converted link should not be replaced twice).

szymonkups:

Fix: Caption will not be automatically added second time if it was already added before "caption fixer" got fired. Closes ckeditor/ckeditor5#5082.

---

### Additional information
Please close after: https://github.com/ckeditor/ckeditor5-image/pull/81

### DIFF
--- a/src/imagecaption/imagecaptionengine.js
+++ b/src/imagecaption/imagecaptionengine.js
@@ -139,7 +139,11 @@ function insertMissingModelCaptionElement( evt, changeType, data, batch ) {
 
 		if ( value.type == 'elementStart' && isImage( item ) && !getCaptionFromImage( item ) ) {
 			batch.document.enqueueChanges( () => {
-				batch.insert( ModelPosition.createAt( item, 'end' ), new ModelElement( 'caption' ) );
+				// Make sure that the image does not have caption already.
+				// https://github.com/ckeditor/ckeditor5-image/issues/78
+				if ( !getCaptionFromImage( item ) ) {
+					batch.insert( ModelPosition.createAt( item, 'end' ), new ModelElement( 'caption' ) );
+				}
 			} );
 		}
 	}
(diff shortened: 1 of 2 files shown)


scofalik:
The fix was okay, but I've changed the test a bit to better reflect original issue.

PR #97, opened 2017-04-12
Introduced standalone view-to-model converter for img element.

### LINKED ISSUE #8: Allow bare <img> to be converted to model
Right now conversion for from view in data pipeline is very strict and only converts images represented by:
```html
<figure class="image">
    <img src="..." alt="..." />
</figure>
```
We should allow single `` (not wrapped in a `<figure>` element) to be converted to model image.

### LINKED ISSUE #5032: Allow bare <img> to be converted to model
Right now conversion for from view in data pipeline is very strict and only converts images represented by:
```html
<figure class="image">
    <img src="..." alt="..." />
</figure>
```
We should allow single `` (not wrapped in a `<figure>` element) to be converted to model image.

scofalik:

Feature: Introduced view-to-model conversion for bare `` elements. Closes ckeditor/ckeditor5#5032.

### DIFF
--- a/src/image/imageengine.js
+++ b/src/image/imageengine.js
@@ -9,8 +9,10 @@
 
 import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
 import buildModelConverter from '@ckeditor/ckeditor5-engine/src/conversion/buildmodelconverter';
-import { viewToModelImage, createImageAttributeConverter } from './converters';
+import buildViewConverter from '@ckeditor/ckeditor5-engine/src/conversion/buildviewconverter';
+import { viewFigureToModel, createImageAttributeConverter } from './converters';
 import { toImageWidget } from './utils';
+import ModelElement from '@ckeditor/ckeditor5-engine/src/model/element';
 import ViewContainerElement from '@ckeditor/ckeditor5-engine/src/view/containerelement';
 import ViewEmptyElement from '@ckeditor/ckeditor5-engine/src/view/emptyelement';
 
@@ -52,8 +54,18 @@ export default class ImageEngine extends Plugin {
 		createImageAttributeConverter( [ editing.modelToView, data.modelToView ], 'src' );
 		createImageAttributeConverter( [ editing.modelToView, data.modelToView ], 'alt' );
 
+		// Build converter for view img element to model image element.
+		buildViewConverter().for( data.viewToModel )
+			.from( { name: 'img', attribute: { src: /.+/ } } )
+			.toElement( ( viewImage ) => new ModelElement( 'image', { src: viewImage.getAttribute( 'src' ) } ) );
+
+		// Build converter for alt attribute.
+		buildViewConverter().for( data.viewToModel )
+			.fromAttribute( 'alt' )
+			.toAttribute( 'alt' );
+
 		// Converter for figure element from view to model.
-		data.viewToModel.on( 'element:figure', viewToModelImage() );
+		data.viewToModel.on( 'element:figure', viewFigureToModel() );
 	}
 }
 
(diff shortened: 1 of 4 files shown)


scofalik:
Will need a PR in engine...

scofalik:
Mentioned PR https://github.com/ckeditor/ckeditor5-engine/issues/919

src/image/imageengine.js
```diff
@@ -52,8 +54,18 @@ export default class ImageEngine extends Plugin {
 		createImageAttributeConverter( [ editing.modelToView, data.modelToView ], 'src' );
 		createImageAttributeConverter( [ editing.modelToView, data.modelToView ], 'alt' );
 
+		// Build converter for view img element to model image element.
+		buildViewConverter().for( data.viewToModel )
+			.from( { name: 'img', attribute: { src: /.+/ } } )
```
Reinmar: Why the `src` check?

scofalik: `src` attribute is required to correctly convert view `` to model `<image>` (which requires `src` attribute). If I understand the code correctly :), this was also implemented in "old" converter (before changes introduced in this PR): https://github.com/ckeditor/ckeditor5-image/pull/97/files#diff-ac5209f1bc4dd7be6f3619d8632013c9L43 here and below.

scofalik:
Testes passes after related PR in engine has been merged.

src/image/imageengine.js
Reinmar: So, being picky, I should just mention that `.` doesn't match new lines.

Reinmar: `/[\s\S]+/` is what we usually use. On the other hand, you didn't add `^` and `$` so, in fact, if that regexp is later used with `match()`, then `.` alone would be enough :D

scofalik: > being picky

Being PK-y?

Reinmar: 😆 

Do you want to change this regexp to `/./`)? Or is that too PK-y?

scofalik: `/./` is fine, but let's please close this already.

Reinmar:
Two small things:

* let's have the alt converter targeting images only,
* and let's have a test what happens when there are two images in a figure (just out of curiosity :D).

scofalik:
Fixed

src/image/imageengine.js
```diff
@@ -57,12 +57,13 @@ export default class ImageEngine extends Plugin {
 		// Build converter for view img element to model image element.
 		buildViewConverter().for( data.viewToModel )
 			.from( { name: 'img', attribute: { src: /./ } } )
-			.toElement( viewImage => new ModelElement( 'image', { src: viewImage.getAttribute( 'src' ) } ) );
+			.toElement( ( viewImage ) => new ModelElement( 'image', { src: viewImage.getAttribute( 'src' ) } ) );
```
Reinmar: You know I was removing these parenthesis in the previous commit? :D

tests/image/imageengine.js
```diff
@@ -147,6 +147,28 @@ describe( 'ImageEngine', () => {
 				expect( getModelData( document, { withoutSelection: true } ) )
 					.to.equal( '<image alt="alt text" src="foo.png"></image>' );
 			} );
+
+			it( 'should not convert alt attribute on non-img element', () => {
```
Reinmar: This test doesn't check anything. Check out src to the previous commit and it will also pass.

tests/image/imageengine.js
```diff
@@ -147,6 +147,28 @@ describe( 'ImageEngine', () => {
...
+				const data = editor.data;
+				const editing = editor.editing;
+
+				document.schema.registerItem( 'div', '$block' );
```
Reinmar, line 155: Maybe it's missing alt allowed here?

src/image/imageengine.js
scofalik: Current standard is having parenthesis around arguments of every arrow function.

tests/image/imageengine.js
scofalik: Well, two things:
1. Allowing alt on div is required, that's true. Without this the test doesn't make much sense.
2. It will pass on previous version, because that was a feature of previous version too.

PR #107, opened 2017-04-26
Enable <img> inside <p> conversion

### LINKED ISSUE #98: Enable <img> inside <p> conversion
Originally described here: https://github.com/ckeditor/ckeditor5-image/issues/8#issuecomment-263228358

The problem is that `<paragraph>` do not allow `<image>` inside it, but it's common to have `<p>Foobar</p>` content in pasted data. Such data should be converted to `<paragraph>Foo</paragraph><image></image><paragraph>bar</p>`.

### LINKED ISSUE #5092: Enable <img> inside <p> conversion
Originally described here: https://github.com/ckeditor/ckeditor5-image/issues/8#issuecomment-263228358

The problem is that `<paragraph>` do not allow `<image>` inside it, but it's common to have `<p>Foobar</p>` content in pasted data. Such data should be converted to `<paragraph>Foo</paragraph><image></image><paragraph>bar</p>`.

scofalik:

Feature: `` view elements placed inside `<p>` elements are now converted to a proper model structure both on `setData()` and pasting (instead of being cleaned). Closes ckeditor/ckeditor5#5092.

---

### Additional information

This implements "autohoisting" idea, that is breaking ancestor(s) of an element if that element would be allowed in context without that ancestor(s). The idea is described in original issue ckeditor/ckeditor5#5092 and also here: https://github.com/ckeditor/ckeditor5-engine/issues/924

Because there were some conceptual problems with introducing generic autohoisting mechanism, I implemented autohoisting just for images for now.


tests/image/imageengine.js
```diff
@@ -168,7 +168,111 @@ describe( 'ImageEngine', () => {
...
+					editor.setData( '<div>foo<img src="foo.jpg" alt="foo" />bar</div>' );
+
+					expect( getModelData( document, { withoutSelection: true } ) ).to.equal(
+						'<div>foo</div>' +
+						'<image alt="foo" src="foo.jpg"></image>' +
+						'<div>bar</div>'
+					);
+				} );
+
+				it( 'multiple images', () => {
```
Reinmar, line 195: Uh... I looked for that test being afraid that you haven't thought about it :D 🥇

Reinmar:
I have simplified bits of your PR. I'm a bit sad that it raises the complexity of the conversion in this feature even more but we have to live with this for a moment. At least, we'll have a lot of material to review when reviewing conversion after 1.0.0 alpha :D

Reinmar:
Make sure to read the commit message that I wrote. I focused on a bit different level – hopefully, more digestible for the readers.

PR #120, opened 2017-06-20
Basic responsive image support.

### LINKED ISSUE #2: Responsive images support
Should image feature be responsible for supporting responsive images? Should this be an editor responsibility?

### LINKED ISSUE #5027: Responsive images support
Should image feature be responsible for supporting responsive images? Should this be an editor responsibility?

szymonkups:

Feature: Support for responsive image's `srcset` attribute. Closes ckeditor/ckeditor5#5027.

### DIFF
--- a/src/image/converters.js
+++ b/src/image/converters.js
@@ -69,34 +69,45 @@ export function viewFigureToModel() {
  *
  * @param {Array.<module:engine/conversion/modelconversiondispatcher~ModelConversionDispatcher>} dispatchers
  * @param {String} attributeName
+ * @param {function} [conversionCallback] Function that will be called each time given attribute conversion is performed.
+ * It will be called with two params: a view image element and type of the conversion: 'addAttribute`, `changeAttribute` or
+ * `removeAttribute`. This callback can be used to perform additional processing on view image element.
  */
-export function createImageAttributeConverter( dispatchers, attributeName ) {
+export function createImageAttributeConverter( dispatchers, attributeName, conversionCallback ) {
 	for ( const dispatcher of dispatchers ) {
-		dispatcher.on( `addAttribute:${ attributeName }:image`, modelToViewAttributeConverter );
-		dispatcher.on( `changeAttribute:${ attributeName }:image`, modelToViewAttributeConverter );
-		dispatcher.on( `removeAttribute:${ attributeName }:image`, modelToViewAttributeConverter );
+		dispatcher.on( `addAttribute:${ attributeName }:image`, modelToViewAttributeConverter( conversionCallback ) );
+		dispatcher.on( `changeAttribute:${ attributeName }:image`, modelToViewAttributeConverter( conversionCallback ) );
+		dispatcher.on( `removeAttribute:${ attributeName }:image`, modelToViewAttributeConverter( conversionCallback ) );
 	}
 }
 
-// Model to view image converter converting given attribute, and adding it to `img` element nested inside `figure` element.
+// Returns model to view image converter converting given attribute, and adding it to `img` element nested inside `figure` element.
+// Allows to provide `conversionCallback` called each time conversion is made.
 //
 // @private
-function modelToViewAttributeConverter( evt, data, consumable, conversionApi ) {
-	const parts = evt.name.split( ':' );
-	const consumableType = parts[ 0 ] + ':' + parts[ 1 ];
+function modelToViewAttributeConverter( conversionCallback ) {
+	return ( evt, data, consumable, conversionApi ) => {
+		const parts = evt.name.split( ':' );
+		const consumableType = parts[ 0 ] + ':' + parts[ 1 ];
 
-	if ( !consumable.consume( data.item, consumableType ) ) {
-		return;
-	}
+		if ( !consumable.consume( data.item, consumableType ) ) {
+			return;
+		}
 
-	const figure = conversionApi.mapper.toViewElement( data.item );
-	const img = figure.getChild( 0 );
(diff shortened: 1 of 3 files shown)


src/image/imageengine.js
```diff
@@ -54,6 +55,20 @@ export default class ImageEngine extends Plugin {
 		createImageAttributeConverter( [ editing.modelToView, data.modelToView ], 'src' );
 		createImageAttributeConverter( [ editing.modelToView, data.modelToView ], 'alt' );
 
+		// Responsive sizes default value. More information and discussion on this topic:
+		// https://github.com/ckeditor/ckeditor5-image/issues/2.
+		config.define( 'image.responsiveSizes', '100vw' );
```
Reinmar: The final decision was – no config option for now: https://github.com/ckeditor/ckeditor5-image/issues/2#issuecomment-309458878.

src/image/imageengine.js
```diff
@@ -70,6 +85,12 @@ export default class ImageEngine extends Plugin {
 			.consuming( { attribute: [ 'alt' ] } )
 			.toAttribute( viewImage => ( { key: 'alt', value: viewImage.getAttribute( 'alt' ) } ) );
 
+		// Build converter for srcset attribute.
+		buildViewConverter().for( data.viewToModel )
+			.from( { name: 'img', attribute: { srcset: /./ } } )
```
Reinmar, line 85: Is this regexp really required by the builder? If it's meant to be a wildcard, it isn't (`.` doesn't match new lines) and matching just one character looks a bit odd in general.

tests/image/imageengine.js
```diff
@@ -47,6 +47,16 @@ describe( 'ImageEngine', () => {
 
 				expect( editor.getData() ).to.equal( '<figure class="image"><img src="foo.png"></figure>' );
 			} );
+
+			it( 'should convert with srcset attribute and add sizes attribute', () => {
+				setModelData( document, '<image src="foo.png" alt="alt text" srcset="small 148w, big.png 1024w"></image>' );
```
Reinmar: Just to be clear – was it intentional that `small` has no extension?

If it's meant to test something, it'd be better if that was a separate test. But I guess that it's just a typo, in  which case it'd be better to add the extension.

Reinmar:
😢 for the changes in the `converters.js`... We must for sure improve the conversion API as one of the first things after the alpha.

src/image/imageengine.js
szymonkups: Ok, so I misunderstood. I thought that we are not going with config option to insert  `[RESPONSIVE_IMAGE_SIZES]` as `sizes` value but staying with possibility to change default value.

src/image/imageengine.js
szymonkups: This object is passed down to the `ViewMatcher`. Matcher will test this regexp against all `img` elements with `srcset` attribute, so it will be true for any non-empty attribute (only one character is needed in this situation to pass `regexp.test()`). 

It looks odd to me too to be honest and I needed to look inside `ViewMatcher` to be sure. Maybe matcher could accept also an array of attributes which will mean "match if attribute is present".

Reinmar: OK, I just wanted to be sure. Thanks. I'll report a ticket to look into this.

Reinmar: https://github.com/ckeditor/ckeditor5-engine/issues/978

src/image/converters.js
```diff
@@ -69,34 +69,45 @@ export function viewFigureToModel() {
...
-	} else {
-		img.setAttribute( data.attributeKey, data.attributeNewValue );
-	}
+		if ( type == 'removeAttribute' ) {
+			img.removeAttribute( data.attributeKey );
+		} else {
+			img.setAttribute( data.attributeKey, data.attributeNewValue );
+		}
+
+		if ( conversionCallback ) {
```
Reinmar, line 107: How did you cover the `else` path? AFAICS this is a private function and it's always called with `conversionCallback`. Do we need this condition at all?

Reinmar: Aaa.. Ok, but `createImageAttributeConverter` is a function too and this param comes from outside.

tests/image/imageengine.js
```diff
@@ -47,6 +47,16 @@ describe( 'ImageEngine', () => {
 
 				expect( editor.getData() ).to.equal( '<figure class="image"><img src="foo.png"></figure>' );
 			} );
+
+			it( 'should convert with srcset attribute and add sizes attribute', () => {
+				setModelData( document, '<image src="foo.png" alt="alt text" srcset="small.png 148w, big.png 1024w"></image>' );
+
+				expect( editor.getData() ).to.equal(
+					'<figure class="image">' +
+						'<img srcset="small.png 148w, big.png 1024w" sizes="100vw" alt="alt text" src="foo.png">' +
```
Reinmar, line 56: This order of attributes looks really bad. Especially that below it's completely different. We need to introduce sorting because such tests have a great chance to blow up on different browsers.

Reinmar: https://github.com/ckeditor/ckeditor5-engine/issues/979

Reinmar:
OK, I have no further comments. Based on the tests I think we can tell that it works, but we'll know more once the upload feature starts using it.

Reinmar:
I also tested pasting. I used https://jsfiddle.net/adamh/ab80gkjz/ and all worked. But I also noticed that we may need to support the `<picture>` notation one day: http://jsfiddle.net/FiNGAHOLiC/fvgf3f8d/ (but I don't really know what it changes).

PR #124, opened 2017-07-06
Added read-only mode to the `ImageTextAlternative`.

### LINKED ISSUE #123: Text alternative balloon should be displayed as read-only when command is disabled
When `ImageTextAlternativeCommand` is disabled user should have access to the alternative text but should not be able to modify it.

### LINKED ISSUE #5108: Text alternative balloon should be displayed as read-only when command is disabled
When `ImageTextAlternativeCommand` is disabled user should have access to the alternative text but should not be able to modify it.

oskarwrobel:

Feature: Added read-only mode to the `ImageTextAlternative`. Closes ckeditor/ckeditor5#5108.

---

Requires https://github.com/ckeditor/ckeditor5-ui/pull/267.


src/imagetextalternative.js
```diff
@@ -98,10 +95,14 @@ export default class ImageTextAlternative extends Plugin {
 	 */
 	_createBalloonPanel() {
 		const editor = this.editor;
+		const command = editor.commands.get( 'imageTextAlternative' );
 
 		const panel = new ImageBalloonPanel( editor );
 		const form = new TextAlternativeFormView( editor.locale );
 
+		form.labeledInput.inputView.bind( 'isReadOnly' ).to( command, 'isEnabled', value => !value );
+		form.saveButtonView.bind( 'isEnabled' ).to( command );
```
oleq: Shouldn't it rather be `form.bind( 'isEnabled' ).to( command );` and then, inside `TextAlternativeFormView`, `isEnabled` should be propagated among the inputs and buttons?

oleq: This way the form will be re–usable with all its features. In your implementation, if you want to re–use the form, you got to repeat the code from the lines above each time.

oleq: If the form had 15 controls, this would become a DRY madness.

oskarwrobel: That was my first thought too. I'll change it.

oskarwrobel: I remember why I didn't do it. Because LinkFormView is bind to two commands `LinkCommand` and `UnlinkCommand`. But image alt use only one command so can have a common TextAlternativeFormView#isEnabled state.

oskarwrobel: I'm not sure about it. `TextAlternativeFormView` is not a reusable view.

oskarwrobel: Done.

oskarwrobel: Maybe we should create some generic FormView? This view might know its children and easily disable and enable it.

oskarwrobel:
I agree with https://github.com/ckeditor/ckeditor5-image/issues/123#issuecomment-313639614. Text alternative is not something that needs to be available in read-only mode. I'm for closing this PR.

PR #131, opened 2017-08-02
Set high priority to prevent ContextualToolbar from being displayed.

### LINKED ISSUE #130: Stopping Contextualtoolbar should have high priority
Now priority is default (normal) and it works (not always) because CT shows after some time, not in the same cycle as ImageToolbar.

### LINKED ISSUE #5112: Stopping Contextualtoolbar should have high priority
Now priority is default (normal) and it works (not always) because CT shows after some time, not in the same cycle as ImageToolbar.

oskarwrobel:

Internal: Set high priority to prevent `ContextualToolbar` from being displayed. Closes ckeditor/ckeditor5#5112.

### DIFF
--- a/src/image.js
+++ b/src/image.js
@@ -54,7 +54,7 @@ export default class Image extends Plugin {
 				if ( selectedElement && isImageWidget( selectedElement ) ) {
 					evt.stop();
 				}
-			} );
+			}, { priority: 'high' } );
 		}
 	}
 }
--- a/tests/image.js
+++ b/tests/image.js
@@ -55,23 +55,35 @@ describe( 'Image', () => {
 		expect( editor.plugins.get( ImageTextAlternative ) ).to.instanceOf( ImageTextAlternative );
 	} );
 
-	it( 'should prevent the ContextualToolbar from being displayed when an image is selected', () => {
-		return ClassicTestEditor.create( editorElement, {
-			plugins: [ Image, ContextualToolbar, Paragraph ]
-		} )
-		.then( newEditor => {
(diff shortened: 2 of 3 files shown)


src/image.js
```diff
@@ -54,7 +54,7 @@ export default class Image extends Plugin {
 				if ( selectedElement && isImageWidget( selectedElement ) ) {
 					evt.stop();
 				}
-			} );
+			}, { priority: 'high' } );
```
oleq, line 57: Can we have some test for it? Is it feasible?

oskarwrobel: I think it is feasible. I'll write one.

oskarwrobel: Done.

tests/image.js
```diff
@@ -87,11 +99,22 @@ describe( 'Image', () => {
 
 			// ContextualToolbar should not be visible.
 			expect( balloon.visibleView ).to.be.null;
+		} );
 
-			// Cleaning up.
-			editorElement.remove();
+		it( 'should listen to ContextualToolbar#show event with high priority', () => {
```
oleq: Not sure if this test proves anything. It passes regardless of the priority of the listener in `Image.init()`, also on master. 

This assertion will always pass
```js
sinon.assert.notCalled( normalPrioritySpy );
```
because on master, the `Image.init()` listener is also attached with the `normal` priority and the order of `listenTo()` calls decides that the one in the test is always the last to execute. In fact – it never gets executed, because `Image` stops the event first.

oskarwrobel: I will add also listener on `high` priority.

PR #136, opened 2017-09-04
Using model selection for calculating caption visibility

### LINKED ISSUE #127: updateCaptionVisibility throws when typing at the end of a link
jul-23-2017 22-58-13

I've got no idea why it throws but this happens in the situation I recorded. For some reason `viewSelection.editableElement` is undefined at that moment.

### LINKED ISSUE #5107: updateCaptionVisibility throws when typing at the end of a link
jul-23-2017 22-58-13

I've got no idea why it throws but this happens in the situation I recorded. For some reason `viewSelection.editableElement` is undefined at that moment.

szymonkups:

Fix: Fixed a bug causing editor with `ImageCaption` plugin enabled threw an error after view got rendered. Closes ckeditor/ckeditor5#5107.

---

### Additional information
View selection was used before it was properly refreshed. This caused errors when trying to access current selection position. It is safer to use model selection in this case.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "@ckeditor/ckeditor5-widget": "^0.2.0"
   },
   "devDependencies": {
+    "@ckeditor/ckeditor5-basic-styles": "^0.9.0",
     "@ckeditor/ckeditor5-clipboard": "^0.7.0",
     "@ckeditor/ckeditor5-dev-lint": "^3.1.0",
     "@ckeditor/ckeditor5-editor-classic": "^0.8.0",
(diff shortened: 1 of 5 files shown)


scofalik:
I propose more informative name for changelog entry.

Instead of
Fix: Using model selection when calculating caption visibility before rendering. Closes #127.

I propose
Fix: Fixed a bug causing editor with `ImageCaption` plugin enabled threw an error after view got rendered. Closes #127.

I'd explain what caused the bug (1-2 sentences) and how it was fixed (1 sentence) in additional information. @Reinmar ?

scofalik:
I miss some kind of test. I know that writing this kinds of tests is a bit frustrating, but if you would be able to reproduce the bug in unit test and then wrap it in `expect( ... ).not.to.throw()` block, it would be perfect.

szymonkups:
Updated PR description by adding "Additional information", added manual test checking the issue.

Reinmar:
> View selection was used before it was properly refreshed. This caused errors when trying to access current selection position. It is safer to use model selection in this case

But why is the view selection outdated? And which part of it, actually?

Reinmar:
>  scofalik merged commit 6147fee into master  26 seconds ago

I guess I'll never know :D

Reinmar:
OK, @scofalik has explained it to me that some nodes are removed from the view but the selection is still anchored in them, so you can't get an editableElement at that moment.

This is surprising and pity. And something which we might need to address anyway. It may cause a lot of bugs that selection is temporarily invalid and some well known properties are nulls.

PR #138, opened 2017-09-08
Hide ImageTextAlternative balloon when image element has been removed

### LINKED ISSUE #137: Image balloon is not removed together with image on external changes
sep-08-2017 16-58-19

### LINKED ISSUE #5115: Image balloon is not removed together with image on external changes
sep-08-2017 16-58-19

oskarwrobel:

Fix: Hide `ImageTextAlternative` balloon when image element has been removed by external changes. Closes ckeditor/ckeditor5#5115.

### DIFF
--- a/tests/image/utils.js
+++ b/tests/image/utils.js
@@ -4,8 +4,11 @@
  */
 
 import ViewElement from '@ckeditor/ckeditor5-engine/src/view/element';
+import ViewSelection from '@ckeditor/ckeditor5-engine/src/view/selection';
+import ViewDocumentFragment from '@ckeditor/ckeditor5-engine/src/view/documentfragment';
+import ViewRange from '@ckeditor/ckeditor5-engine/src/view/range';
 import ModelElement from '@ckeditor/ckeditor5-engine/src/model/element';
-import { toImageWidget, isImageWidget, isImage } from '../../src/image/utils';
+import { toImageWidget, isImageWidget, isImageWidgetSelected, isImage } from '../../src/image/utils';
 import { isWidget, getLabel } from '@ckeditor/ckeditor5-widget/src/utils';
 
 describe( 'image widget utils', () => {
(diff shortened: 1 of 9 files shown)


tests/image/utils.js
```diff
@@ -49,6 +52,40 @@ describe( 'image widget utils', () => {
...
+
+			const selection = new ViewSelection( [ ViewRange.createOn( notWidgetizedElement ) ] );
+
+			expect( isImageWidgetSelected( selection ) ).to.be.false;
+		} );
+
+		it( 'should return false when widget element is not the only element in the selection', () => {
+			const notWidgetizedElement = new ViewElement( 'p' );
+
+			frag = new ViewDocumentFragment( [ notWidgetizedElement, notWidgetizedElement ] );
```
oleq: Shouldn't this be `new ViewDocumentFragment( [ element, notWidgetizedElement ] )` instead?

oleq:
Shouldn't this issue be included in http://localhost:8125/ckeditor5-ui/tests/manual/contextualballoon/externalchanges.html or provide a new MT?

oleq (changes requested):
Minor changes needed in tests.

oskarwrobel:
> Shouldn't this issue be included in http://localhost:8125/ckeditor5-ui/tests/manual/contextualballoon/externalchanges.html or provide a new MT?

I was thinking about it but this is something that we are able to test by our unit tests. I think MT is useful to test repositioning because this is something we need to stub in unit tests.

But I can add MT if you think is needed for this case.

oleq:
For one thing, I was surprised that the MT is in ckeditor5-ui instead of ckeditor5-link. The code which handles re-positioning and stuff on `#render` isn't in the `ContextualBalloon`. It's a pure link stuff. 

Then I thought that we should check the same thing in the image and I finally found out there's a corresponding test in the image too ckeditor5-image/tests/manual/tickets/106/1.html. It doesn't test this kind of issue because what is "remotely" removed is the paragraph before the image instead of the image widget itself.

So maybe, let's just extend this MT and include this issue's case?

oskarwrobel:
I was using `106` MT to prepare the GIF for this issue. OK, I'll extend this test.

oleq:
The MT is OK. But it revealed that the editor loses focus when the image which text alternative is being edited gets removed. This, I'm afraid, must be addressed (it [works](https://github.com/ckeditor/ckeditor5-link/blob/master/src/link.js#L246-L248) in the link feature because it explicitly calls focus).

PR #140, opened 2017-09-13
t/134: Various improvements to config.image.styles

### LINKED ISSUE #134: Make it possible to align images in builds with little effort
One of the common requirements for end users will be the possibility of aligning images in the way how people got used to it: to the left/right side in the same article (instead of being able to use just a side image aligned to one side, offered by default).

The current method of defining such styles is far from being ideal, requiring the developer to provide complex configuration and build own editor, thus making it impossible to use CKEditor 5 from CDN.

```
import fullSizeIcon from '@ckeditor/ckeditor5-core/theme/icons/object-center.svg';
import alignLeftIcon from '@ckeditor/ckeditor5-core/theme/icons/object-left.svg';
import alignRightIcon from '@ckeditor/ckeditor5-core/theme/icons/object-right.svg';

...
            styles: [
                // This option is equal to a situation where no style is applied.
                {
                    name: 'imageStyleFull',
                    title: 'Full size image',
                    icon: fullSizeIcon,
                    value: null
                },

                // This represents an image aligned to left.
                {
                    name: 'imageStyleLeft',
                    title: 'Left aligned image',
                    icon: alignLeftIcon,
                    value: 'left',
                    className: 'image-style-left'
                },

                // This represents an image aligned to right.
                {
                    name: 'imageStyleRight',
                    title: 'Right aligned image',
                    icon: alignRightIcon,
                    value: 'right',
                    className: 'image-style-right'
                }
```

Even if there was some trick/workaround to provide icons (proposed in https://github.com/ckeditor/ckeditor5-ui/issues/293) I believe that does not solve the problem of being cumbersome. Here's why:
  - One has to copy & paste some long blob of text to achieve a common setup.
  - One has to point to some external icons.
  - One has to define a title, which will not be translated out of the box to the large number of languages supported by CKEditor 5.

What I'd love to see is the the possibility of using some simplified configuration to let the Image plugin know which image styles are available. E.g. using only names of styles that CKEditor 5 would then recognise:

```
styles : [ { name: 'imageStyleLeft' } , { name: 'imageStyleFull' } , { name: 'imageStyleRight' } ]

if one wishes to use different class names:

styles : [ { name: 'imageStyleLeft',  className: 'mystyle-left' } , ... ]
```
CKEditor 5 could use some sane defaults for the rest of properties that were not set (e.g. icons, titles translated on Transifex). To support passing such a simplified configuration, the Image plugin would have to come with two additional "built-in" styles, which are not enabled by default 
 ('imageStyleLeft'/'imageStyleRight') but at least possible to enable in an easy way.

Note: the syntax above is just to explain the idea of not having to define an icon and the the whole styles left/right from scratch. There may be better ways of setting this up easily.

### LINKED ISSUE #135: Changing the orientation of the side image icon
It may be a bit related to #134.

What if a developer sets the style of .image-style-side to `float: left;`. Is there a way to change the sideIcon from `object-right.svg` to `object-left.svg` when using a CKEditor build?

### LINKED ISSUE #5113: Make it possible to align images in builds with little effort
One of the common requirements for end users will be the possibility of aligning images in the way how people got used to it: to the left/right side in the same article (instead of being able to use just a side image aligned to one side, offered by default).

The current method of defining such styles is far from being ideal, requiring the developer to provide complex configuration and build own editor, thus making it impossible to use CKEditor 5 from CDN.

```
import fullSizeIcon from '@ckeditor/ckeditor5-core/theme/icons/object-center.svg';
import alignLeftIcon from '@ckeditor/ckeditor5-core/theme/icons/object-left.svg';
import alignRightIcon from '@ckeditor/ckeditor5-core/theme/icons/object-right.svg';

...
            styles: [
                // This option is equal to a situation where no style is applied.
                {
                    name: 'imageStyleFull',
                    title: 'Full size image',
                    icon: fullSizeIcon,
                    value: null
                },

                // This represents an image aligned to left.
                {
                    name: 'imageStyleLeft',
                    title: 'Left aligned image',
                    icon: alignLeftIcon,
                    value: 'left',
                    className: 'image-style-left'
                },

                // This represents an image aligned to right.
                {
                    name: 'imageStyleRight',
                    title: 'Right aligned image',
                    icon: alignRightIcon,
                    value: 'right',
                    className: 'image-style-right'
                }
```

Even if there was some trick/workaround to provide icons (proposed in https://github.com/ckeditor/ckeditor5-ui/issues/293) I believe that does not solve the problem of being cumbersome. Here's why:
  - One has to copy & paste some long blob of text to achieve a common setup.
  - One has to point to some external icons.
  - One has to define a title, which will not be translated out of the box to the large number of languages supported by CKEditor 5.

What I'd love to see is the the possibility of using some simplified configuration to let the Image plugin know which image styles are available. E.g. using only names of styles that CKEditor 5 would then recognise:

```
styles : [ { name: 'imageStyleLeft' } , { name: 'imageStyleFull' } , { name: 'imageStyleRight' } ]

if one wishes to use different class names:

styles : [ { name: 'imageStyleLeft',  className: 'mystyle-left' } , ... ]
```
CKEditor 5 could use some sane defaults for the rest of properties that were not set (e.g. icons, titles translated on Transifex). To support passing such a simplified configuration, the Image plugin would have to come with two additional "built-in" styles, which are not enabled by default 
 ('imageStyleLeft'/'imageStyleRight') but at least possible to enable in an easy way.

Note: the syntax above is just to explain the idea of not having to define an icon and the the whole styles left/right from scratch. There may be better ways of setting this up easily.

### LINKED ISSUE #5114: Changing the orientation of the side image icon
It may be a bit related to ckeditor/ckeditor5-image#134.

What if a developer sets the style of .image-style-side to `float: left;`. Is there a way to change the sideIcon from `object-right.svg` to `object-left.svg` when using a CKEditor build?

oleq:

Feature: Allowed customization of default image styles. Defined formatting–oriented styles. Simplified `config.image.styles` syntax. Closes ckeditor/ckeditor5#5113. Closes ckeditor/ckeditor5#5114.

BREAKING CHANGE: From now on, the `imageStyleFull` uses `object-full-width.svg` icon.
 
---

### Additional information

Requires:
* https://github.com/ckeditor/ckeditor5-core/tree/t/ckeditor5-image/134 (additional icon).
* https://github.com/ckeditor/ckeditor5/tree/t/ckeditor5-image/134 (docs and CI)

Showcase in http://localhost:8125/ckeditor5-image/tests/manual/imagestyle.html.

---

Default config:
```js
image: {
	styles: [ 'imageStyleFull', 'imageStyleSide' ]
}
```

Customizing just the icon and class in defaults:
```
image: {
	styles: [ 
		{ name: 'imageStyleFull', icon: 'left' }, 
		{ name: 'imageStyleSide', className: 'foo' } 
	]
}
```

Completely custom styles also benefit from translations and default icons:
```js
image: {
	styles: [
		// This style's title will be translated, e.g. to german. It uses default right icon.
		{ name: 'foo', title: 'Full size image', icon: 'right', value: null, className: 'foo-class' },
		// This style does everything in a custom way.
		{ name: 'bar', title: 'bar', icon: '<svg>...</svg>', value: 'bar', className: 'bar-class' },
	]
}
```


wwalc:
Any chance the default CSS rules for `imageStyleAlignLeft` could play better by default with the content inside CKEditor? I used the default browser styles and rendered Classic Editor inside. I used blockquote and bulleted list and both look bad:

### Align right (ok)

screen shot 2017-09-14 at 09 18 32

### Align left (could be better)

screen shot 2017-09-14 at 09 18 41

Not sure if it's an issue with ckeditor5-image or related plugins responsible for blockquote/lists.
However anyone using these styles will face that problem so it would be nice to come with some solution.

oleq:
For one thing, I could increase the [default margin](https://github.com/ckeditor/ckeditor5-image/blob/t/134/theme/theme.scss#L6) around the images.

Then there's the problem with lists and blockquotes that we faced recently. It boils down to `overflow: hidden` for `<ol>` and `<blockquote>` to enable [block formatting](https://stackoverflow.com/a/6128264/1485219). CKE4 is suffering because of that too (and the quote most of all):

image

Reinmar:
I'm not sure how far we can go. We can increase margins – that's fine. But can we use `overflow:hidden`? That has, as we learnt, a pretty unclear effect. Developers may be confused by this.

OTOH, I'd love if the initial styling was perfect...

oleq:
> OTOH, I'd love if the initial styling was perfect...

I'm afraid there's nothing we can do. It's overflow, or nothing :/

image

image


There's a long discussion on this topic https://stackoverflow.com/questions/710158/why-do-my-list-item-bullets-overlap-floating-elements.

Reinmar:
Hm... OK, seeing this I'm fine with overflow :D

fredck:
You mean fine, as long as the text of long paragraphs wrap around the image, right?

wwalc:
**TL;DR:**

I'd like to propose the following:

* Increase the right margin for `.ck-editor__editable .image.image-style-align-left`
* Apply overflow: hidden for blockquotes.

Reasoning below.

----

I guess there is no perfect solution. _Thinking out loud:_ in terms of frequency of using semantic block structures I guess the order will be like:
 * paragraphs
 * headings
 * lists
 * ... long break ...
 * quotes

## Paragraphs, headings
There is no issue to solve.

## Lists

### Bigger margin

`.ck-editor__editable .image.image-style-align-left {margin-right: 2em}`

screen shot 2017-09-15 at 10 30 33

Pros: the bullets are visible.
Pros: plays well with long lists.
Pros: the heading has better margin, IMO.

### Overflow: hidden

screen shot 2017-09-15 at 10 30 11

Pros: the bullets are visible.
Cons: I don't like the significant difference between heading and lists margin.  
Cons: in case of very long lists looks weird. Personally I create long lists quite often.

**IMO: bigger margin wins**

## Quotes

### No changes

screen shot 2017-09-15 at 10 41 51


### Bigger margin (to solve issue with lists)
screen shot 2017-09-15 at 10 38 39

### Overflow: hidden + bigger margin (to solve issue with lists)

screen shot 2017-09-15 at 10 38 23

Pros: in this case at least shorter quotes look okay. If someone wants to use longer quotes he will have to use either centered/full size image or align it to right, there is no better solution.

**IMO: Overflow: hidden + bigger margin wins.**

Reinmar:
Agree. I didn't think that overflow:hidden will make long list indented on the entire height of them. This is bad for lists, but makes sense for quotes.

wwalc:
It looks like defining a completely custom style does not work, the `className` is not respected.

```
ClassicEditor
  .create( document.querySelector( '#editor' ), {
    image: {
      toolbar: [ 'imageTextAlternative', '|', 'imageStyleCustom', 'imageStyleAlignCenter', 'imageStyleAlignRight' ],

      styles: [
         // THIS ONE DOES NOT WORK
         { name: 'imageStyleCustom', icon: 'left', className: 'foo-class', title: 'Custom title' },
         'imageStyleAlignCenter',
         { name: 'imageStyleAlignRight', className: 'bar-class' }
      ]
    }
  })
.catch( error => {
  console.error( error );
} );
```

`imageStyleCustom` correctly shows up in the image toolbar, but the class is not applied. Note that overriding className for `imageStyleAlignRight` works.

oleq:
@wwalc  Does it show up in `editor.plugins.get( ImageStyleEngine ).imageStyles` (the class)?

wwalc:
@oleq  I modified the manual test in http://localhost:8125/ckeditor5-image/tests/manual/imagestyle.html

```
ClassicEditor
	.create( document.querySelector( '#editor-semantic' ), {
		plugins: [
			ImageToolbar,
			EnterPlugin,
			TypingPlugin,
			ParagraphPlugin,
			HeadingPlugin,
			ImagePlugin,
			UndoPlugin,
			ClipboardPlugin,
			ImageStyle
		],
		toolbar: [ 'headings', 'undo', 'redo' ],
		image: {
			// You need to configure the image toolbar too, so it uses the new style buttons.
			toolbar: [ 'imageTextAlternative', '|', 'imageStyleCustom', 'imageStyleAlignCenter', 'imageStyleAlignRight' ],

			styles: [
				{ name: 'imageStyleCustom', icon: 'left', className: 'foo-class', title: 'Custom title' },
				'imageStyleAlignCenter',
				{ name: 'imageStyleAlignRight', className: 'bar-class' }
			]
		}
	} )
	.then( editor => {
		window.editorSemantic = editor;
	} )
	.catch( err => {
		console.error( err.stack );
	} );
```

With such test, in Firefox in the console I have the following results:

screen shot 2017-09-15 at 18 06 33

Note: I initially reproduced the same issue in a manually created classic build. So I can confirm it using both a build and a modified manual test.

oleq:
@wwalc 

```js
{ name: 'imageStyleCustom', icon: 'left', className: 'foo-class', title: 'Custom title' },
```
does not work because it has no `value` defined. I'm wondering if we should `log.warn` in such case 🤔  Or at least make it clear in the [documentation](https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/api/module_image_imagestyle_imagestyleengine-ImageStyleFormat.html).

oleq:
I increased the default image margins for better interaction with lists. An overflow for blockquote should be added in a followup in the ckeditor5-block-quote repository.

BTW, I found a funny bug in Chrome. Can you also reproduce it? I hope it's not CKEditor–specific.

kapture 2017-09-19 at 13 43 02

wwalc:
> does not work because it has no value defined. I'm wondering if we should log.warn in such case 🤔 Or at least make it clear in the documentation.

Doh. It's fine. I have no idea why I missed the value property. I must have copied something that did not contain `value` defined at all. 

BTW. Wouldn't it be more intuitive is the default `value` was set to whatever the `className` is set? Or to something like `class-className`? After all when defining styles I wouldn't care most of the time about some internal model attribute. Also isn't there a chance that when asking users to define manually such thing they may break something else (e.g. using `title`, `alt` or whatever as a value)?

oleq:
TBH, I also find it a little bit confusing that the `value` must always be provided as a `null|String`. 

Do you remember why we decided to go this way, @szymonkups? Can we do anything about this?

szymonkups:
> Do you remember why we decided to go this way, @szymonkups? Can we do anything about this?

This is implemented this way just to provide a way to configure model value for the style. Looking at this now I think @wwalc opinion has more sense to treat it as some internals that should be hidden from the configurator. Value is stored inside known attribute so I think that it is safe to create value based on style name and/or class name. This will require to mark default style in some different way - currently by setting `null` as the `value`.

Reinmar:
> This will require to mark default style in some different way - currently by setting null as the value.

Yup. Good point.

Agree with the rest too.

src/imagestyle/converters.js
```diff
@@ -48,7 +48,7 @@ export function modelToViewStyleAttribute( styles ) {
  */
 export function viewToModelStyleAttribute( styles ) {
 	// Convert only styles without `null` value.
```
wwalc: Is the comment still valid here?

oleq:
I got rid of `ImageStyleFormat#value` in https://github.com/ckeditor/ckeditor5-image/pull/140/commits/ad1a540a6a097b292a46944b4779a779da95ed88. Please let me know if it works for you and does not break anything.

It also means some fixes in documentation are needed in the ckeditor5 repo https://github.com/ckeditor/ckeditor5/tree/t/ckeditor5-image/134.

I also enabled CI in https://github.com/ckeditor/ckeditor5/tree/t/ckeditor5-image/134.

wwalc (approved):
I did not check the whole PR deeply, but the things that I requested to be corrected seem to work okay now. I'll review and update all the image docs once this PR merged.

oleq:
@szymonkups Could you review this issue?

Reinmar:
I'd start from the fact that CI for this branch fails :P

Reinmar:
Oopsie – I haven't noticed that this feature requires changes in other pkgs.

PR #169, opened 2018-01-29
Removed image autohoisting.

oskarwrobel:

Internal: Removed autohoisting converters in favor of conversion builder. Closes https://github.com/ckeditor/ckeditor5-engine/issues/1213.

---

Part of: https://github.com/ckeditor/ckeditor5-engine/pull/1256


src/imagestyle/converters.js
```diff
@@ -49,41 +47,27 @@ export function viewToModelStyleAttribute( styles ) {
...
-	if ( !consumable.test( viewFigureElement, { class: style.className } ) ) {
-		return;
-	}
 
-	// Check if figure is converted to image.
-	if ( !isImage( modelImageElement ) ) {
-		return;
-	}
+		const viewFigureElement = data.input;
+		const modelImageElement = data.output.getItems().next().value;
```
pjasiun: Since `writer.setAttribute` already accepts range, we should make `schema.checkAttribute` also accepts range and we will be able to get rid off this ugly line. In fact, it is schema context what should accept range. It should build the context based on the first element in that range or throw an exception if there are more elements.

pjasiun: Or you can use `first( output.getItems() );` ;)

oskarwrobel: For now I'll change it to `first()`. I'm already using `first()` in other image converters to get range first item.

PR #177, opened 2018-02-14
Align feature class naming to a new scheme.

jodator:

Other: Align feature class naming to a new scheme.

BREAKING CHANGE: Renamed `ImageEngine` to `ImageEditing`.
BREAKING CHANGE: Extracted `ImageUI` from `Image`.
BREAKING CHANGE: Renamed `ImageStyleEngine` to `ImageStyleEditing`.
BREAKING CHANGE: Extracted `ImageStyleUI` from `ImageStyle`.
BREAKING CHANGE: Renamed `ImageTextAlternativeEngine` to `ImageTextAlternativeEditing`.
BREAKING CHANGE: Extracted `ImageTextAlternativeUI` from `ImageTextAlternative`.

---

### Additional information

* Part of feature split: https://github.com/ckeditor/ckeditor5/issues/488

### DIFF
--- a/docs/features/image.md
+++ b/docs/features/image.md
@@ -148,7 +148,7 @@ See the result below:
 
 ### Defining custom styles
 
-Besides using the {@link module:image/imagestyle/imagestyleengine~ImageStyleEngine.defaultStyles 5 predefined styles}:
+Besides using the {@link module:image/imagestyle/imagestyleediting~ImageStyleEditing.defaultStyles 5 predefined styles}:
 
 * `'imageStyleFull'`,
 * `'imageStyleSide'`,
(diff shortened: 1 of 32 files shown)


scofalik:
I might have messed up something during merge with master, but here are my doubts:

https://github.com/ckeditor/ckeditor5-image/blob/t/ckeditor5/488/src/image.js#L50-L56

Is this supposed to be here? Doesn't it touch UI? Shouldn't it be somewhere where image toolbar is declared (`imagetoolbar.js`)? _(I moved it to `imagetoolbar.js` and with a little fix all tests pass)_

https://github.com/ckeditor/ckeditor5-image/blob/t/ckeditor5/488/src/imagestyle/imagestyleediting.js

I don't know if I like that there is some UI stuff here. For example: icons. Shouldn't it be moved to `imagestyleui.js`?

PR #183, opened 2018-03-08
 Other: Update naming of UI components & commands.

jodator:

Other: Update naming of UI components & commands.

BREAKING CHANGE: Renamed `imageUpload` command to `uploadImage`.
BREAKING CHANGE: The "imageStyleFull", "imageStyleSide", "imageStyleAlignLeft", "imageStyleAlignRight" and "imageStyleAlignCenter" commands are no longer available. Replaced by "imageStyle" command that accepts name of configured image style as a value.
BREAKING CHANGE: The "imageStyleFull", "imageStyleSide", "imageStyleAlignLeft", "imageStyleAlignRight" and "imageStyleAlignCenter" UI components are no longer available. Replaced by "imageStyle:full", "imageStyle:side", "imageStyle:alignLeft", "imageStyle:alignRight" and "imageStyle:alignCenter".
BREAKING CHANGE: The `ImageStyleCommand#value` is no longer boolean only. Now it can have a name of image style of current image element.
BREAKING CHANGE: The `ImageStyleCommand` constructor's second parameter is now an array of supported model attribute values.

---

### Additional information

* PR of : https://github.com/ckeditor/ckeditor5-alignment/pull/25.

### DIFF
--- a/docs/_snippets/features/image-style-custom.js
+++ b/docs/_snippets/features/image-style-custom.js
@@ -12,16 +12,16 @@ ClassicEditor
 		image: {
 			styles: [
 				// This option is equal to a situation where no style is applied.
-				'imageStyleFull',
+				'imageStyle:full',
 
 				// This represents an image aligned to left.
-				'imageStyleAlignLeft',
+				'imageStyle:alignLeft',
 
 				// This represents an image aligned to right.
-				'imageStyleAlignRight'
+				'imageStyle:alignRight'
 			],
 
-			toolbar: [ 'imageTextAlternative', '|', 'imageStyleAlignLeft', 'imageStyleFull', 'imageStyleAlignRight' ]
+			toolbar: [ 'imageTextAlternative', '|', 'imageStyle:alignLeft', 'imageStyle:full', 'imageStyle:alignRight' ]
 		},
 		toolbar: {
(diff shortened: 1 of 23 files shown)


docs/_snippets/features/image-style-custom.js
```diff
@@ -12,16 +12,16 @@ ClassicEditor
 		image: {
 			styles: [
 				// This option is equal to a situation where no style is applied.
-				'imageStyleFull',
+				'imageStyle:full',
```
Reinmar: It doesn't make sense to repeat the `imageStyle:` prefix in `image.styles` config. I'll remove it.

src/imagestyle.js
```diff
@@ -42,7 +42,7 @@ export default class ImageStyle extends Plugin {
  * The default value is:
  *
  *		const imageConfig = {
- *			styles: [ 'imageStyleFull', 'imageStyleSide' ]
+ *			styles: [ 'imageStyle:full', 'imageStyle:side' ]
```
Reinmar: Well, this is odd. Below the name is set to `full` and here you use `imageStyle:full`. So I think that you wanted to change this but forgot at some point :D

Reinmar:
Also, the bit about the commands is wrong:

image

PR #201, opened 2018-04-04
Image element should be cleared from progress if uploadId changed to null

### LINKED ISSUE #200: ImageUploadProgress: image element should be "cleaned" when it loses `uploadId`
At this moment, if in the same batch, `uploadId` is changed to `null`, the `uploadStatus` change is not converted, because `uploadId` is already `null` and the converter returns quickly.

https://github.com/ckeditor/ckeditor5-image/blob/master/src/imageupload/imageuploadprogress.js#L61-L65

Instead, I think that if `uploadId` is `null`, then image should be cleared out of anything connected with progress. Then, it will work correctly if multiple deltas will come together from collaboration.

### LINKED ISSUE #5145: ImageUploadProgress: image element should be "cleaned" when it loses `uploadId`
At this moment, if in the same batch, `uploadId` is changed to `null`, the `uploadStatus` change is not converted, because `uploadId` is already `null` and the converter returns quickly.

https://github.com/ckeditor/ckeditor5-image/blob/master/src/imageupload/imageuploadprogress.js#L61-L65

Instead, I think that if `uploadId` is `null`, then image should be cleared out of anything connected with progress. Then, it will work correctly if multiple deltas will come together from collaboration.

scofalik:
Fix: Image element will be cleared from upload progress classes if `uploadId` attribute changed to `null`. Closes ckeditor/ckeditor5#5145.

### DIFF
--- a/src/imageupload/imageuploadprogress.js
+++ b/src/imageupload/imageuploadprogress.js
@@ -60,12 +60,12 @@ export default class ImageUploadProgress extends Plugin {
 		const modelImage = data.item;
 		const uploadId = modelImage.getAttribute( 'uploadId' );
 
-		if ( !conversionApi.consumable.consume( data.item, evt.name ) || !uploadId ) {
+		if ( !conversionApi.consumable.consume( data.item, evt.name ) ) {
 			return;
 		}
 
 		const fileRepository = editor.plugins.get( FileRepository );
-		const status = data.attributeNewValue;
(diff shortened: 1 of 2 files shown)


scofalik:
I've rebased with `master`. Two tests fail, but they fail on `master` too.

pjasiun (approved):

PR #205, opened 2018-05-14
Feature: Added upload complete icon.

### LINKED ISSUE #204: Add confirmation check mark to the image upload animation
There should be :white_check_mark: when the upload finish sucessfully.

This is how it looks now in CKEditor 5:

before

This is how the animation looks like in Letters (with check mark):

after

Without it, upload looks incomplete.

### LINKED ISSUE #5147: Add confirmation check mark to the image upload animation
There should be :white_check_mark: when the upload finish sucessfully.

This is how it looks now in CKEditor 5:

before

This is how the animation looks like in Letters (with check mark):

after

Without it, upload looks incomplete.

oskarwrobel:

Feature: Added upload complete icon. Closes ckeditor/ckeditor5#5147.

### DIFF
--- a/src/imageupload/imageuploadprogress.js
+++ b/src/imageupload/imageuploadprogress.js
@@ -10,6 +10,7 @@
 import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
 import FileRepository from '@ckeditor/ckeditor5-upload/src/filerepository';
 import uploadingPlaceholder from '../../theme/icons/image_placeholder.svg';
+import UIElement from '@ckeditor/ckeditor5-engine/src/view/uielement';
 import ViewPosition from '@ckeditor/ckeditor5-engine/src/view/position';
 import ViewRange from '@ckeditor/ckeditor5-engine/src/view/range';
 
@@ -98,6 +99,10 @@ export default class ImageUploadProgress extends Plugin {
 			}
 
 			return;
+		}
+
+		if ( status == 'complete' ) {
+			_showCompleteIcon( viewFigure, viewWriter, editor.editing.view );
 		}
 
 		// Clean up.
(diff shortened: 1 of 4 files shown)


src/imageupload/imageuploadprogress.js
```diff
@@ -193,6 +201,21 @@ function _hideProgressBar( viewFigure, writer ) {
...
+//
+// @param {module:engine/view/containerelement~ContainerElement} viewFigure
+// @param {module:engine/view/writer~Writer} writer
+// @param {module:engine/view/view~View} view
+function _showCompleteIcon( viewFigure, writer, view ) {
+	const completeIcon = new UIElement( 'div', { class: 'ck-image-upload-finish' } );
+
+	writer.insert( ViewPosition.createAt( viewFigure, 'end' ), completeIcon );
+
+	setTimeout( () => {
```
oskarwrobel: https://github.com/ckeditor/ckeditor5-image/issues/206

Mgsy:
The animation doesn't display properly in Edge.

bug_cke5

dkonopka:
@mgsy ☝️ It's related to `calc()` in the animation https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/12872907/, anyways even with "hard-coded" values, there are some issues with `transform-origin`.

I'm not so happy about this, but probably we should drop this animation in check icon..

theme/imageuploadicon.css
```diff
@@ -0,0 +1,92 @@
...
+
+:root {
+	--ck-image-upload-icon-size: 1.5em;
+	--ck-image-upload-icon-width: 2px;
+
+	--ck-color-image-upload-icon: hsl(0, 0%, 100%);
+	--ck-color-image-upload-icon-background: hsl(120, 100%, 27%);
+}
+
+.ck-image-upload-complete-icon {
```
oleq: I see the wireframe styles (display, position) mixed up with pure visual styles (background, animation) here. The later should be moved to Lark.

Mgsy (approved):
Everything works fine.

pjasiun:
I also see that green check is bigger and is shown longer than before. It looks much heavier than before. I remember that we had the discussion that is should not be there at all, that the users should have a feeling that the image is loaded in the first place, so the subtle confirmation is a compromise which takes some time. Can we only bring what we had in Letters in this ticket (the same size and time, maybe different color), and move the discussion about changing it to a separate ticket, if needed?

PR #212, opened 2018-06-22
Support uploading images pasted from other websites

### LINKED ISSUE #211: Inconsistent image pasting
When copy-pasting a local image into the editor, it will be uploaded as expected.

However, if you copy an image from another website and paste it into the editor, it will use the previous url as the source, rather than uploading it.

Is this the intended behaviour? This can sometimes cause issues with regards to hotlinking, dead links,  security, etc.

### LINKED ISSUE #5152: Handle image pasted/dropped from various sources (external websites, base64-encoded, etc.)
When copy-pasting a local image into the editor, it will be uploaded as expected.

However, if you copy an image from another website and paste it into the editor, it will use the previous url as the source, rather than uploading it.

Is this the intended behaviour? This can sometimes cause issues with regards to hotlinking, dead links,  security, etc.

mjadobson:
Previous check prevented uploads when copy pasting images from other websites, as these drag events include html data in addition to file data.

Closes ckeditor/ckeditor5#5152 .

---

Image uploads are now triggered when copy pasting an image from a website.

However, they will still not be triggered when drag + dropping an image from a website. This is because the drag event doesn't include file data, but `text/uri-list` data, which includes the image url(s). Similarly, images won't be uploaded if selecting and copying+dragging/pasting a bunch of html.

In these instances, due to potential CORS issues, you would need to have a server endpoint that downloads and saves images directly from these external urls.

Additionally, it may also be worthwhile having a config option to validate img src, to only insert them when the url matches certain patterns. This could be used to prevent re-uploading previously uploaded content, and restrict hotlinking from other websites.

### DIFF
--- a/src/imageupload/imageuploadediting.js
+++ b/src/imageupload/imageuploadediting.js
@@ -47,9 +47,9 @@ export default class ImageUploadEditing extends Plugin {
 
 		// Execute imageUpload command when image is dropped or pasted.
 		this.listenTo( editor.editing.view.document, 'clipboardInput', ( evt, data ) => {
-			// Skip if non empty HTML data is included.
+			// Skip if no files to upload
 			// https://github.com/ckeditor/ckeditor5-upload/issues/68
-			if ( isHtmlIncluded( data.dataTransfer ) ) {
+			if ( !data.dataTransfer.files || !data.dataTransfer.files.length ) {
 				return;
 			}
 
@@ -230,12 +230,4 @@ export default class ImageUploadEditing extends Plugin {
 			fileRepository.destroyLoader( loader );
 		}
 	}
-}
-
-// Returns `true` if non-empty `text/html` is included in the data transfer.
-//
-// @param {module:clipboard/datatransfer~DataTransfer} dataTransfer
-// @returns {Boolean}
-export function isHtmlIncluded( dataTransfer ) {
-	return Array.from( dataTransfer.types ).includes( 'text/html' ) && dataTransfer.getData( 'text/html' ) !== '';
 }
--- a/tests/imageupload/imageuploadediting.js
+++ b/tests/imageupload/imageuploadediting.js
@@ -199,23 +199,6 @@ describe( 'ImageUploadEditing', () => {
 		expect( getModelData( model ) ).to.equal( '<paragraph>foo[]</paragraph>' );
 	} );
 
-	it( 'should not insert image when there is non-empty HTML content pasted', () => {
-		const fileMock = createNativeFileMock();
-		const dataTransfer = new DataTransfer( {
-			files: [ fileMock ],
-			types: [ 'Files', 'text/html' ],
-			getData: type => type === 'text/html' ? '<p>SomeData</p>' : ''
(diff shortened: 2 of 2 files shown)


Reinmar:
Thanks for the PR! As the first thing to do, I replied to your ticket: https://github.com/ckeditor/ckeditor5-image/issues/211#issuecomment-399440794

I haven't checked your changes yet, but maybe my answer will clarify why the current implementation looks like this. We'll also have to think what changes we can do to it without hitting the need to make this configurable.

Reinmar:
One quick question (before I'll pass this to someone who worked on the upload feature): what will happen if in the clipboard there will be HTML data + files? I think about this scenario:

> HTML with images which src are references to files in the clipboard's dataTransfer.files,

cc @mlewand, do I remember correctly that in some cases we got something like this in CKEditor 4?

mlewand:
@Reinmar TBH I'm not aware of a way to link/reference data from your dataTransfer in `text/html` content. The only exception is Safari, which inlines images pasted from Word. It does that by replacing `img[src]` value with `blob://xyz` (I'm not 100% positive whether it was blob protocol, talking from the top of my head, better confirm that).

Reinmar:
> @Reinmar TBH I'm not aware of a way to link/reference data from your dataTransfer in text/html content. The only exception is Safari, which inlines images pasted from Word. It does that by replacing img[src] value with blob://xyz (I'm not 100% positive whether it was blob protocol, talking from the top of my head, better confirm that).

So maybe I had this in my mind. I haven't worked with clipboard for some time now :) Thanks.

Reinmar:
@mjadobson, I realised that a complete solution to https://github.com/ckeditor/ckeditor5-image/issues/211 may be beyond our capabilities now, so I went back to check your PR. 

The most important thing that we miss here is tests. Your PR is supposed to add handling for files copied from other websites, yes? We need tests which simulate such situations. We can't merge a PR without that. From these tests I'd like to learn what kind of HTML is in the clipboard when you copy images from other websites. Is this `` tag?

Another thing is that your PR changes the logic that files in the `dataTransfer` have precedence over any HTML. This will completely break pasting if any of the browsers happen to put both at the same time – e.g. a content from MSWord and files in that content. We need to be extremely careful here and perhaps add an additional check – if there's `text/html` in the clipboard, checking what it contains.

bkalafat:
How to use ImageResize with react. I can't find any sample. 

```
    <CKEditor editor={ClassicEditor}
          data="<p>Hello from CKEditor 5!</p>"
          onInit={editor => {
            editor.plugins.get(
              "FileRepository"
            ).createUploadAdapter = loader => {
              return new UploadAdapter(loader)
            }

            editor.plugins.get("ImageResize")

            console.log("Editor is ready to use!", editor)
          }}
          onChange={(event, editor) => {
            const data = editor.getData()
            console.log({ event, editor, data })
          }}
          onBlur={(event, editor) => {
            console.log("Blur.", editor)
          }}
          onFocus={(event, editor) => {
            console.log("Focus.", editor)
          }}
    />
```

PR #215, opened 2018-06-26
Used UI#update event instead of View#render to attach the UI components.

### LINKED ISSUE #142: Image should re-render view on load

### LINKED ISSUE #5118: Image should re-render view on load

oskarwrobel:

Internal: Used UI#update event instead of View#render to attach the UI components. Closes ckeditor/ckeditor5#5118.

---

### Part of

- https://github.com/ckeditor/ckeditor5-core/pull/131


oskarwrobel:
Interesting 🤔 

image

PR #218, opened 2018-07-05
Added integration tests for ckeditor/ckeditor5#1120

pomek:

Tests: Added integration tests for ckeditor/ckeditor5#1120.

### DIFF
--- a/tests/imagecaption/imagecaption-integration.js
+++ b/tests/imagecaption/imagecaption-integration.js
@@ -27,7 +27,7 @@ describe( 'ImageCaption integration', () => {
 		editorElement.remove();
 	} );
 
-	describe( 'Enter plugin', () => {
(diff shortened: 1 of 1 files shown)


Reinmar:
This test doesn't fail even when https://github.com/ckeditor/ckeditor5/issues/1120 was still open because it checks how enter commands change the model. It does not check what would the browser do if you pressed <kbd>Enter</kbd> because you cannot recreate that. You can only check if `preventDefault()` was called in such tests.

Other than that, there's little point in checking model and view structures. Check one – usually the model.

Reinmar (changes requested):
As commented.

pomek:
Ok, fixed. I've checked whether these tests fail on commit before fixing the bug. They failed.

PR #219, opened 2018-07-05
Added CSS upload loader

### LINKED ISSUE #207: Consider changing image upload placeholder
At this moment we are using [svg as placeholder](https://github.com/ckeditor/ckeditor5-image/blob/master/theme/icons/image_placeholder.svg) with plain text `Uploading image...` and progress bar animation. The problem is there is no way to translate this text, because it's hardcoded in the `svg` file. 

upload-placeholder

We should think about something like centered spinner instead of plain text.

cc @oleq @Reinmar

### LINKED ISSUE #5150: Consider changing image upload placeholder
At this moment we are using [svg as placeholder](https://github.com/ckeditor/ckeditor5-image/blob/master/theme/icons/image_placeholder.svg) with plain text `Uploading image...` and progress bar animation. The problem is there is no way to translate this text, because it's hardcoded in the `svg` file. 

upload-placeholder

We should think about something like centered spinner instead of plain text.

cc @oleq @Reinmar

oskarwrobel:

Feature: Added CSS upload loader. Closes ckeditor/ckeditor5#5150.

---

### Additional information

**Required [ckeditor5-theme-lark/t/ckeditor5-image/207](https://github.com/ckeditor/ckeditor5-theme-lark/tree/t/ckeditor5-image/207)**

Hardcoded text `Uploading image...` has been replaced by a nice looking CSS spinner.

SVG file is still used as an image source while the upload is in progress because:
1. image can't have an empty source
2. svg file sets the size of the image and this size is scalable (e.g. when switching image to side image)

### Preview 

upload-placeholder


pjasiun (approved):

oskarwrobel:
@dkonopka 

image
https://github.com/ckeditor/ckeditor5-image/pull/219/files#diff-1dd60618d87727a52cc7ebc65fbaa2eeR26

image

Mgsy (approved):

theme/imageuploadloader.css
```diff
@@ -0,0 +1,26 @@
...
+	justify-content: center;
+	width: 100%;
+	height: 100%;
+	top: 0;
+	left: 0;
+
+	&::before {
+		content: '';
+		position: relative;
+		width: var(--ck-upload-placeholder-loader-size);
```
oleq: Styles like `width`, `height`, belong to Lark. `position`, `top`, `content`, `flex` should stay here.

oleq (changes requested):

theme/icons/image_placeholder.svg
```diff
@@ -1 +1 @@
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 700 250"><g fill="none" fill-rule="evenodd"><rect width="700" height="250" fill="#FAFAFA" rx="4"/><text fill="#5F6F77" font-family="Arial,sans-serif" font-size="24"><tspan x="247.9" y="135">Uploading image…</tspan></text></g></svg>
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 700 250"><g fill="#FAFAFA" fill-rule="evenodd"><rect width="700" height="250" rx="4"/></g></svg>
```
oleq, line 1: BTW, why do we need that SVG? It's displaying a gray area, that's what it does? Couldn't we use a CSS element instead?

oskarwrobel: > SVG file is still used as an image source while the upload is in progress because:

> 1. image can't have an empty source
> 2. svg file sets the size of the image and this size is scalable (e.g. when switching image to side image)

oleq:
I can see 2 scenarios:

1. I upload the image
   * the horizontal bar on the top is displaying the progress,
   * I can see the image immediately so there's no need to display the spinner
2. Someone else is uploading
   * the horizontal bar on the top is moving,
   * I can see the spinner until the actual image comes from the server

The first scenario looks great:

image

But the later looks as follows

kapture 2018-07-09 at 15 39 02

and I find this double indicator an overkill. Too much movement, pure redundancy. Let's stick to the spinner in that situation, WDYT?

cc @dkonopka @oskarwrobel @pjasiun

dkonopka:
☝️ Good catch!

"Infinite progress top bar" should be removed if someone else is uploading image, I just completely forgot to get rid of it.

pjasiun:
I also did notice it and agree that 2 "spinners" (center and top) is redundancy. But it is not a big issue for me.

oskarwrobel:
> I upload the image
> - the horizontal bar on the top is displaying the progress,
> - I can see the image immediately so there's no need to display the spinner

The spinner is displayed during `reading` state. When the image is huge then `FileReader` may need a while to load this image.

Issue #247, opened 2018-11-14
[Edge] Support for uploading images with base64 source


f1ames:
This is a follow up #246.

Since the [`File API`](https://developer.mozilla.org/en-US/docs/Web/API/File) is used during [image base64 to file conversion](https://github.com/ckeditor/ckeditor5-image/blob/t/246/src/imageupload/imageuploadediting.js#L327-L336), it doesn't work on Edge as `File` constructor is not supported there (see https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/9551546/).

Alternatively the `Blob` object could be used, however blobs itself do not provide a way to define a filename. This is problematic since during upload the filename is read from the file object and send with other data:

![image](https://user-images.githubusercontent.com/1061942/48478166-c73cda00-e803-11e8-85bc-ff8d49e71b99.png)

and with `Blob` instance, the filename is always set to "blob":

![image](https://user-images.githubusercontent.com/1061942/48478297-2a2e7100-e804-11e8-88f2-136e4f036f29.png)

which doesn't work well with upload integration:

* File with the same name is always send to the server.
* File name doesn't have the extension so checking file type by extension fails.

---

We need to figure out a way to provide custom filename so it can be send with upload request. It may be done by adjusting blob object so the browser uses custom name which was added (not 100% sure, my initial tests failed here). The second option is to pass filename to a loader so it is used further in the uploading process.

f1ames:
During development of the main feature (#246), we have decided that all uploaded images will have `image.ext` name, which means that server-side implementation of the file uploading should take care of proper naming.
The only problem with Edge is that files are uploaded with `blob` filename - but apart from that `Blob`'s can be uploaded the same way as `File`'s. So maybe if we had assumed that server-side code takes care of naming, we could upload the blob files in Edge (which seems to be better solution than just removing images from pasted content) and add a proper documentation for this case for handling `blob` named files. WDYT? cc @Reinmar

PR #248, opened 2018-11-14
Support for uploading images with base64 source

### LINKED ISSUE #246: Support for uploading images with base64 source
When the image with base64 `src` is inserted or an image `src` attribute is modified to base64, upload should be triggered and image uploaded to a server (via available adapter).

This will partially cover #21 and solve https://github.com/ckeditor/ckeditor5-paste-from-office/issues/24.

### LINKED ISSUE #2524: Pasted images should trigger upload if Image Upload plugin is enabled
Follow up of ckeditor/ckeditor5#2515.

When content containing images is pasted from Word (or any other supported Office suite) the image upload should be triggered so pasted images could be automatically uploaded to a server.

### LINKED ISSUE #5172: Support for uploading images with base64 source
When the image with base64 `src` is inserted or an image `src` attribute is modified to base64, upload should be triggered and image uploaded to a server (via available adapter).

This will partially cover https://github.com/ckeditor/ckeditor5-paste-from-office/issues/21 and solve https://github.com/ckeditor/ckeditor5-paste-from-office/issues/24.

f1ames:

Feature: Support for uploading images with base64 source. Closes ckeditor/ckeditor5#5172. Closes ckeditor/ckeditor5#2524.

---

### Additional information

See ckeditor/ckeditor5#5172 (and https://github.com/ckeditor/ckeditor5-paste-from-office/issues/24#issuecomment-436186985 which explains the approach a little). ATM Edge browser is not supported, see ckeditor/ckeditor5#5173.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "ckeditor5-plugin"
   ],
   "dependencies": {
+    "@ckeditor/ckeditor5-clipboard": "^10.0.3",
     "@ckeditor/ckeditor5-core": "^11.0.1",
     "@ckeditor/ckeditor5-engine": "^11.0.0",
     "@ckeditor/ckeditor5-theme-lark": "^11.1.0",
@@ -20,7 +21,6 @@
   },
   "devDependencies": {
     "@ckeditor/ckeditor5-basic-styles": "^10.0.3",
-    "@ckeditor/ckeditor5-clipboard": "^10.0.3",
     "@ckeditor/ckeditor5-editor-classic": "^11.0.1",
     "@ckeditor/ckeditor5-enter": "^10.1.2",
     "@ckeditor/ckeditor5-essentials": "^10.1.2",
--- a/package.json
+++ b/package.json
@@ -10,7 +10,6 @@
     "ckeditor5-plugin"
   ],
   "dependencies": {
-    "@ckeditor/ckeditor5-clipboard": "^10.0.3",
     "@ckeditor/ckeditor5-core": "^11.0.1",
     "@ckeditor/ckeditor5-engine": "^11.0.0",
     "@ckeditor/ckeditor5-theme-lark": "^11.1.0",
@@ -21,6 +20,7 @@
   },
   "devDependencies": {
     "@ckeditor/ckeditor5-basic-styles": "^10.0.3",
+    "@ckeditor/ckeditor5-clipboard": "^10.0.3",
     "@ckeditor/ckeditor5-editor-classic": "^11.0.1",
     "@ckeditor/ckeditor5-enter": "^10.1.2",
     "@ckeditor/ckeditor5-essentials": "^10.1.2",
--- a/src/imageupload/imageuploadediting.js
+++ b/src/imageupload/imageuploadediting.js
@@ -7,6 +7,8 @@
  * @module image/imageupload/imageuploadediting
  */
 
+/* global fetch, File */
+
 import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
 import FileRepository from '@ckeditor/ckeditor5-upload/src/filerepository';
 import Notification from '@ckeditor/ckeditor5-ui/src/notification/notification';
@@ -72,6 +74,41 @@ export default class ImageUploadEditing extends Plugin {
 					} );
 				}
 			} );
+		} );
+
+		// Handle images inserted or modified with base64 source.
+		doc.on( 'change', () => {
+			const changes = doc.differ.getChanges( { includeChangesInGraveyard: false } );
+			const imagesToUpload = [];
+
+			for ( const entry of changes ) {
+				let item = null;
+
+				if ( entry.type == 'insert' && entry.name == 'image' ) {
+					// Process entry item if it was an image insertion.
+					item = entry.position.nodeAfter;
+				} else if ( entry.type == 'attribute' && entry.attributeKey == 'src' ) {
+					// Process entry item if it was modification of `src` attribute of an image element.
+					// Such cases may happen when image with `blob` source is inserted and then have it
+					// converted to base64 data by clipboard pipeline.
+					const el = entry.range.start.nodeAfter;
+
+					// Check if modified element is an image element.
+					if ( el && el.is( 'image' ) ) {
+						item = el;
+					}
+				}
+
+				if ( item && !item.getAttribute( 'uploadId' ) && item.getAttribute( 'src' ) &&
+					item.getAttribute( 'src' ).match( /data:image\/\w+;base64/ ) ) {
+					imagesToUpload.push( item );
+				}
+			}
+
+			// Upload images with base64 sources.
+			if ( imagesToUpload.length ) {
+				this._uploadBase64Images( imagesToUpload, editor );
+			}
 		} );
 
 		// Prevents from the browser redirecting to the dropped image.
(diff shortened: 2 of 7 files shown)


f1ames:
Ready for review.

f1ames:
I have changed the way upload integrates with base64/blob images. Now `imageuploadediting` listens to `inputTransformation` event and finds all images with base64/blob source. The images sources are fetched as `File` instance (except for Edge, see #247) and then file is passed to a file uploader (from this point it works the same as regular file uploading flow as it reuses its logic too).

**Important changes**:

1. As `inputTransforamtion` is emitted by the `Clipboard` plugin I have added it as a package dependency (not sure if it is a **breaking change** for the package?) Alternatively, if we don't want to add it as a dep, the code could check if `Clipboard` plugin is present and perform image processing only if plugin is available.
2. The base64 source could be transformed to `File` in a synchronous manner, however blob cannot. To not make the code too complicated by using different methods for fetching base64 and blob there is the same async method (using [`Fetch API`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API)) used for both. It also guarantees non-blocking fetching (so UI won't freeze).
3. The above (async fetching) requires `inputTransformation` to be stopped and re-fired after images are fetched.
4. I'm quite sad about the Edge browser, especially that `Blob`'s can be also used for upload same as `File` objects. However, as mentioned in #247, when `Blob` is send, the filename used for the file during upload is always `blob` which may not work with some upload adapters (for example CKFinder doesn't play well with it - it seems it does some type checking based on filename). However, if we decide that we can move the responsibility of proper file naming to adapter or server-side uploading we can go with blobs on Edge. ATM images are simply removed as they cannot be uploaded (no `File` instance available) which is not the best solution (remember we can't keep them with base64 source to not break collaboration).

---

**Slightly different approach for fetching images**:

As mentioned in points **2** and **3** above the flow for fetching base64/blob images is:

1. Listen to `inputTransformation` event.
1. Find images with local sources in clipboard data.
1. If there are any, stop `inputTransformation` event.
1. Fetch images (async) and start upload.
1. Modify clipboard data based on fetch result.
1. Fire `inputTransformation` event.

I was wondering if the flow could be altered as:

1. Listen to `inputTransformation` event.
1. Find images with local sources in clipboard data.
1. Modify images in clipboard data so base64/blob is not propagated to model.
1. Fetch images (async) and start upload.

So in the second variant, the `inputTransformation` event is not paused/delayed by fetching images. It means the pasted content will appear faster in the editor. However, as images are not ready the content will appear without image previews. Preview images will appear when fetched which will result in resizing/moving some parts of the content. This is just an alternative which I was thinking of, not sure if it brings any real improvement, but wanted to mention it.

package.json
```diff
@@ -10,6 +10,7 @@
     "ckeditor5-plugin"
   ],
   "dependencies": {
+    "@ckeditor/ckeditor5-clipboard": "^10.0.3",
```
Reinmar: That's not necessary. You're not using it anyway. You're only using `plugins.get( 'Clipboard' )` which can be satisfied by other implementations of this feature (not only `ckeditor5-clipboard`).

src/imageupload/imageuploadediting.js
```diff
@@ -81,6 +93,54 @@ export default class ImageUploadEditing extends Plugin {
...
+				return;
+			}
+
+			evt.stop();
+
+			Promise.all( fetchableImages ).then( items => {
+				for ( const item of items ) {
+					if ( !item.file ) {
+						// Failed to fetch image or create a file instance, remove image element.
+						view.change( writer => {
```
Reinmar: Do we need to use the change block here? It's upcast writing, isn't it?

src/imageupload/imageuploadediting.js
```diff
@@ -81,6 +93,54 @@ export default class ImageUploadEditing extends Plugin {
...
+
+						if ( loader ) {
+							view.change( writer => {
+								writer.setAttribute( 'src', '', item.image );
+								writer.setAttribute( 'uploadId', loader.id, item.image );
+							} );
+						} else {
+							view.change( writer => {
+								// Set attribute so the image will not be processed 2nd time.
+								writer.setAttribute( 'uploadProcessed', true, item.image );
```
Reinmar: I'd move this out of this condition. It would be more intuitive to read if we marked all processed images. Not only half of them. After all, we don't want to process again all images that we already processed.

src/imageupload/utils.js
```diff
@@ -18,3 +20,78 @@ export function isImageType( file ) {
...
+/**
+ * Creates a promise which fetches the image local source (base64 or blob) and returns as a `File` object.
+ *
+ * @param {module:engine/view/element~Element} image Image which source to fetch.
+ * @param {Number} index Image index used as image name suffix.
+ * @returns {Promise} A promise which resolves when image source is fetched and converted to `File` instance.
+ * It resolves with object holding initial image element (as `image`) and its file source (as `file`). If
+ * the `file` attribute is null, it means fetching failed.
+ */
+export function wrapImageToFetch( image, index ) {
```
Reinmar: `fetchLocalImage()`?

src/imageupload/utils.js
```diff
@@ -18,3 +20,78 @@ export function isImageType( file ) {
...
+/**
+ * Checks whether given node is an image element with local source (base64 or blob).
+ *
+ * @param {module:engine/view/node~Node} node Node to check.
+ * @returns {Boolean}
+ */
+export function isLocalImage( node ) {
+	return node.is( 'element', 'img' ) && node.getAttribute( 'src' ) &&
+		( node.getAttribute( 'src' ).match( /data:image\/\w+;base64,/g ) ||
+		node.getAttribute( 'src' ).match( /blob:/g ) );
```
Reinmar:
```suggestion
			if ( !node.is( 'element', 'img' ) || node.getAttribute( 'src' ) ) {
				return false;
			}
			
			return ( node.getAttribute( 'src' ).match( /data:image\/\w+;base64,/g ) ||
		node.getAttribute( 'src' ).match( /blob:/g ) );
```

Reinmar: Also, I think these URL regexps may need to start with `^`?

src/imageupload/utils.js
```diff
@@ -18,3 +20,78 @@ export function isImageType( file ) {
...
+ */
+export function wrapImageToFetch( image, index ) {
+	return new Promise( resolve => {
+		// Fetch works asynchronously and so does not block browser UI when processing data.
+		fetch( image.getAttribute( 'src' ) )
+			.then( resource => resource.blob() )
+			.then( blob => {
+				const mimeType = getImageMimeType( blob, image.getAttribute( 'src' ) );
+				const ext = mimeType.replace( 'image/', '' );
+				const filename = `${ Number( new Date() ) }-image${ index }.${ ext }`;
```
Reinmar: I think that `image.${ ext }` will be enough.

Reinmar (changes requested):
We can simplify a couple of things here.

package.json
f1ames: I see, wouldn't 

```js
this.listenTo( editor.plugins.get( 'Clipboard' ), ... );
```

fail if there will not be any `Clipboard` feature in the CKEditor build?

f1ames: ☝️ cc @Reinmar

Reinmar: Yes, it would fail. But we do the same thing with other feature-feature relations. This simplifies recomposing the editor with e.g. a bit different implementation of one of those features.

We use `Plugin#requires()` for feature-util and gluefeature-subfeature relations.

PR #254, opened 2018-11-21
fix: handle null file types in image check

ckotzbauer:

fix: handle null file types in image check

---

### Additional information

This PR fixes JS errors if files with null filetypes or null-files are processed by the "isImageType" function. I cannot reproduce such cases locally, but our customers can do this with D&D in the newest Chrome...

### DIFF
--- a/src/imageupload/imageuploadediting.js
+++ b/src/imageupload/imageuploadediting.js
@@ -55,7 +55,13 @@ export default class ImageUploadEditing extends Plugin {
 				return;
 			}
 
-			const images = Array.from( data.dataTransfer.files ).filter( isImageType );
+			const images = Array.from( data.dataTransfer.files ).filter( file => {
+				if ( !file ) {
+					return false;
+				}
+
+				return isImageType( file );
+			} );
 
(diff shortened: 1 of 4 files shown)


Reinmar:
I feel unsafe about this change being made in a util. It may hide some bigger issues if this util will be used in more places. I'd still not be ideal, but I'd move this code directly to the place where `isImageType()` is used. Most likely you mean this place:

https://github.com/ckeditor/ckeditor5-image/blob/25be8e0a84efeebcad1b01905a1153bc5e999a5d/src/imageupload/imageuploadediting.js#L58

Also, this seems to be a workaround for some Chrome bug, so that's why it'd be better to keep this change more locally.

Reinmar:
BTW, any chance that you could find out more about how and when this issue occurs? It's a bit odd one :|

ckotzbauer:
Yes, I mean this place with the datatransfer.
We currently try to find out more about this issue from our users. Hopefully we get some useful infos.

Reinmar:
Do you think you could update the PR this week? We're finishing an iteration now and we would include it in the release.

ckotzbauer:
@Reinmar Done, I moved the null check to the dnd-logic.

Reinmar:
Thanks! It looks great.

PR #262, opened 2018-12-13
Enable image in table cells.

jodator:

Feature: Enable image in table cells.

---

### Additional information

* Part of: https://github.com/ckeditor/ckeditor5-table/pull/155

### DIFF
--- a/src/image/utils.js
+++ b/src/image/utils.js
@@ -100,7 +100,7 @@ export function isImageAllowed( model ) {
 	const schema = model.schema;
 	const selection = model.document.selection;
 
-	return isImageAllowedInParent( selection, schema, model ) && checkSelectionWithObject( selection, schema );
+	return isImageAllowedInParent( selection, schema, model ) && checkSelectionOnObject( selection, schema ) && isInOtherImage( selection );
 }
 
 // Checks if image is allowed by schema in optimal insertion parent.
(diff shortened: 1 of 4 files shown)


J-Nagle:
Thanks very much for responding to our feedback in https://github.com/ckeditor/ckeditor5-table/pull/155#issuecomment-446797016. However, this gave us a `TypeError` when we went to test it. Attached is our build, below is the error, and here are links to the [configuration](https://pastebin.com/6WvB13au) and [stack trace](https://pastebin.com/xaJq9xLn). Hopefully this is enough...

```
TypeError: t.plugins.has is not a function
	at af.init (imageuploadediting.js:94)
InlineEditor.create.then.catch.error @ app.js:335 // When our system calls CKEditor
```
[CKEditor - Table image upload test build.zip](https://github.com/ckeditor/ckeditor5-image/files/2692701/CKEditor.-.Table.image.upload.test.build.zip)

jodator:
@J-Nagle looks like latest changes in master :) I've updated PR (#t/ckeditor5-table/99) branches on
- ckeditor/ckeditor5-engine
- ckeditor/ckeditor5-widget

It should be OK but I'll test it better tomorrow morning as now I'm not able to properly test it.

Also I'll try to push work on this PR faster but nevertheless those changes will be publicly available in next release. Now it may fail as `master` branch is considered a working branch.

J-Nagle:
Thanks for the fast response! The updated masters of Engine and Widget still don't make this work for us right now, but I absolutely appreciate the quick tweak nonetheless. Of course, we understand that other work may take priority, and will definitely upgrade to the next release when available. Cheers!

src/image/utils.js
```diff
@@ -112,29 +112,28 @@ function isImageAllowedInParent( selection, schema, model ) {
...
+// Check if selection is on object.
 //
 // @returns {Boolean}
-function checkSelectionWithObject( selection, schema ) {
+function checkSelectionOnObject( selection, schema ) {
 	const selectedElement = selection.getSelectedElement();
 
-	const isSelectionOnObject = !!selectedElement && schema.isObject( selectedElement );
-	const isSelectionInObject = !![ ...selection.focus.getAncestors() ].find( ancestor => schema.isObject( ancestor ) );
+	return !( !!selectedElement && schema.isObject( selectedElement ) );
```
scofalik: Doesn't this return the opposite thing? If this is checking if the selection is on object it should simply return:

`return selectedElement && schema.isObject( selectedElement );`

src/image/utils.js
```diff
@@ -112,29 +112,28 @@ function isImageAllowedInParent( selection, schema, model ) {
...
+}
 
-	return !isSelectionOnObject && !isSelectionInObject;
+// Checks if selection is placed in other image (ie. in caption).
+function isInOtherImage( selection ) {
+	return [ ...selection.focus.getAncestors() ].every( ancestor => !ancestor.is( 'image' ) );
```
scofalik, line 128: Simple optimization would be to check `.some` xD.

scofalik (changes requested):
Small things.

src/image/utils.js
jodator: How would it be better then `.every`?

Both will return if condition is other then expected:
```js
.some( a => a.is( 'image' ) );
.every( a => !a.is( 'image' ) );
```

PR #258, opened 2018-12-07
Tests: Adjusted tests to changes in upload plugin

f1ames:
### Suggested merge commit message ([convention](https://github.com/ckeditor/ckeditor5-design/wiki/Git-commit-message-convention))

Tests: Adjusted tests to changes in upload plugin.

---

### Additional information

Tests adjusted to changes from https://github.com/ckeditor/ckeditor5-upload/issues/87. Should be closed after or together with https://github.com/ckeditor/ckeditor5-upload/pull/88.

The CI will fail untill https://github.com/ckeditor/ckeditor5-upload/pull/88 will be merged to `master` so it will be good to merge https://github.com/ckeditor/ckeditor5-upload/pull/88 first and then rerun the CI job for this PR to check if it is green.

PR #264, opened 2018-12-17
The image uploading listener for handling `base64/blob` images no longer stops `inputTransformation` event

### LINKED ISSUE #263: The image uploading listener for handling base64/blob images should not stop `inputTransformation` event
Extracted from https://github.com/ckeditor/ckeditor5-paste-from-office/issues/44. See https://github.com/ckeditor/ckeditor5-paste-from-office/issues/44#issuecomment-443200544.

### LINKED ISSUE #2530: [Edge] Nothing is pasted when text + images selected
nov-29-2018 17-21-29

It's strange because it soooometimes work (content appears, although images have local src, but that's a known Edge problem). 

If we can't handle images in this case I'd expect that at least text is pasted.

### LINKED ISSUE #5177: The image uploading listener for handling base64/blob images should not stop `inputTransformation` event
Extracted from https://github.com/ckeditor/ckeditor5-paste-from-office/issues/44. See https://github.com/ckeditor/ckeditor5-paste-from-office/issues/44#issuecomment-443200544.

f1ames:

Other: The image uploading listener for handling `base64/blob` images no longer stops `inputTransformation` event. Closes ckeditor/ckeditor5#5177. Closes ckeditor/ckeditor5#2530.

---

### Additional information

This PR aligns the way `base64/blob` images uploading works to https://github.com/ckeditor/ckeditor5-upload/issues/87 issue in which the `FileLoader` accepts `Promise` instead of `File` instance.

It's nice to see how much `inputTransformation` listener got simplified. However, tests got a little more complex.

_This PR is based on #258 PR and will not work without https://github.com/ckeditor/ckeditor5-upload/issues/88 so should be closed after closing this mentioned PRs. The CI will be red unless https://github.com/ckeditor/ckeditor5-upload/issues/88 is merged._


jodator:
@f1ames I've forgot about this one after checking the images upload. It is still valid right?
If so could you check conflicts with master? I'll review it first thing next business day :) :coffee:

f1ames:
Still failing on Edge, due to some reason 🤔 I will look into it.

src/imageupload/imageuploadediting.js
```diff
@@ -101,36 +101,25 @@ export default class ImageUploadEditing extends Plugin {
...
 					.filter( value => isLocalImage( value.item ) && !value.item.getAttribute( 'uploadProcessed' ) )
-					.map( value => fetchLocalImage( value.item ) );
+					.map( value => { return { promise: fetchLocalImage( value.item ), imageElement: value.item }; } );
 
 				if ( !fetchableImages.length ) {
 					return;
 				}
 
-				evt.stop();
+				const writer = new UpcastWriter();
```
jodator: Hmm I'm worried about that construct. I think that we could expose those methods on the view if they're need. Then we could write it as:

```js
const view = editor.editing.view; // ps.: also used for crateRangeIn()

/// the code as now

view.setAttribute( 'src', '', fetchableImage.imageElement );
```
This unfortunately will require small changes in the engine.

jodator, line 110: Hmm I'm worried about that construct. I think that we could expose those methods on the view if they're need. Tahen we could write it as:

```js
const view = editor.editing.view; // ps.: also used for crateRangeIn()

/// the code as now

view.setAttribute( 'src', '', fetchableImage.imageElement );
```
This unfortunately will require small changes in the engine.

Edit: I've just found this construct in the Clipboard docs. I'll report an issue there. So do not change this ATM.

src/imageupload/utils.js
```diff
@@ -22,29 +22,29 @@ export function isImageType( file ) {
 }
 
 /**
- * Creates a promise which fetches the image local source (base64 or blob) and returns as a `File` object.
+ * Creates a promise which fetches the image local source (base64 or blob) and resolves with a `File` object.
  *
  * @param {module:engine/view/element~Element} image Image which source to fetch.
  * @returns {Promise} A promise which resolves when image source is fetched and converted to `File` instance.
```
jodator:
```suggestion
 * @returns {Promise.<File>} A promise which resolves when image source is fetched and converted to `File` instance.
```

src/imageupload/utils.js
```diff
@@ -22,29 +22,29 @@ export function isImageType( file ) {
...
-				const mimeType = getImageMimeType( blob, image.getAttribute( 'src' ) );
+				const mimeType = getImageMimeType( blob, imageSrc );
 				const ext = mimeType.replace( 'image/', '' );
 				const filename = `image.${ ext }`;
 				const file = createFileFromBlob( blob, filename, mimeType );
 
-				resolve( { image, file } );
+				file ? resolve( file ) : reject();
 			} )
 			.catch( () => {
```
jodator: Meybe leave it as `.catch( reject )`? WDYT will be shorter.

jodator (changes requested):
Cosmetics to change really.

src/imageupload/imageuploadediting.js
jodator: ps.: reporter here: https://github.com/ckeditor/ckeditor5-clipboard/issues/46#issuecomment-454025196

jodator:
As the code is ready: @Mgsy could you check this one?

src/imageupload/imageuploadediting.js
f1ames: Ok, I'll leave it as is for now. Btw. one may use `change` block here:

```js
editor.editing.view.change( writer => {
    // code...
} );
```

to obtain the `writer`. However, it uses `DowncastWriter` not `UpcastWriter` (still their `setAttribute()` methods uses exactly same API).

jodator: I was thinking abut it and while it looks safe we probably ought to fix it globally as in referenced ticket.

Mgsy:
I'm not able to paste an image from Word to the editor in Edge (other browsers work fine and the sample has PFO plugin activated). Tested on VM.

bug_cke5

f1ames:
> I'm not able to paste an image from Word to the editor in Edge (other browsers work fine and the sample has PFO plugin activated). Tested on VM.

That's probably caused by the lack of support for `File` constructor in Edge. However, in this case the image should be pasted (as blob) and no error should be shown (I guess it is not handled correctly ATM).

jodator:
> That's probably caused by the lack of support for `File` constructor in Edge. However, in this case the image should be pasted (as blob) and no error should be shown (I guess it is not handled correctly ATM).

So should this be fixed here or in PFO plugin?

f1ames:
☝️ Sorry, didn't make it clear in the previous comment. It should be fixed in this PR from what I can see and I'm working on the fix ATM.

f1ames:
> However, in this case the image should be pasted (as blob) and no error should be shown (I guess it is not handled correctly ATM).

Mixed up a little bit, we cannot use blob here due to collaboration features (it may break it). So the image simply will not be pasted. However, `undefined` alert is unnecessary. I think we should skip any messages here.

f1ames:
Funny that there is no error in Edge in the case @Mgsy found. In Chrome (when file promise is rejected) there is - so I have added a fix there too - https://github.com/ckeditor/ckeditor5-upload/pull/90. Please review and merge it together with this PR.

Mgsy:
>I have added a fix there too - ckeditor/ckeditor5-upload#90. Please review and merge it together with this PR.

After switching to this branch, indeed, the image isn't pasted in Edge, but unfortunately, the `undefined` alert is still displayed.

f1ames:
@Mgsy I have pushed a fix. In Edge image should not be pasted, no errors in the console and no alert displayed.

> After switching to this branch, indeed, the image isn't pasted in Edge, but unfortunately, the undefined alert is still displayed.

Sorry, it was part of the fix, the second part was pushed in the latest commit (https://github.com/ckeditor/ckeditor5-image/pull/264/commits/f5a52f3eee29cb8aa0b5e8225d526f10b7d78ebc).

Mgsy (approved):
👌

jodator (approved):

PR #278, opened 2019-02-18
T/ckeditor5/1524: Insert missing caption for images that are nested in other elements.

jodator:

Fix: Insert missing caption for images that are nested in other elements. Closes ckeditor/ckeditor5#1524.

---

### Additional information

* I've used the TreeWalker to check inserted content for images. I'm not sure if this is optimal but right now I don't a need to introduce anything new to the differ (ie list of inserted child nodes for give `insert` type of change).

### DIFF
--- a/src/imagecaption/imagecaptionediting.js
+++ b/src/imagecaption/imagecaptionediting.js
@@ -185,40 +185,34 @@ export default class ImageCaptionEditing extends Plugin {
 		let wasFixed = false;
 
 		for ( const entry of changes ) {
-			if ( entry.type == 'insert' ) {
-				const images = [];
+			const images = [];
 
+			if ( entry.type == 'insert' && entry.name != '$text' ) {
 				const item = entry.position.nodeAfter;
 
-				if ( item ) {
-					if ( entry.name == 'image' ) {
-						images.push( item );
-					} else {
-						if ( !item.is( 'image' ) && item.childCount ) {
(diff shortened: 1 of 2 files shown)


src/imagecaption/imagecaptionediting.js
```diff
@@ -182,17 +178,39 @@ export default class ImageCaptionEditing extends Plugin {
...
-				if ( !getCaptionFromImage( item ) ) {
-					writer.appendElement( 'caption', item );
+				if ( item.is( 'image' ) && !getCaptionFromImage( item ) ) {
+					imagesWithoutCaption.push( item );
+				}
+
+				// Check elements with children for nested images.
+				if ( !item.is( 'image' ) && item.childCount ) {
+					// Use the walker to find all nested images despite of their nest level.
+					const walker = model.createRangeOn( item ).getWalker();
```
scofalik: You can use `model.createRangeOn( ... ).getItems()`. It will be easier to iterate through elements in a range.

scofalik: Also, since you already know that `item` is not an image, you can go with `createRangeIn()`. Right?

jodator: Thanks for the tips. I didn't know about `.getItems()` and you're right about `crateRangeIn()` I suppose ;)

Fixing.

PR #280, opened 2019-02-20
Fixed triple click selection in figcaption

pomek:

Fix: Fixed a weird selection behavior on triple click. Triple click inside the `figcaption` will select the entire element now. Closes ckeditor/ckeditor5#1463.


oleq (changes requested):
I'm afraid must fix the root of the problem and this fix should land in the widget plugin. Run http://localhost:8125/ckeditor5-widget/tests/manual/widget-with-nestededitable.html to see that the scope of the issue is wider than just images.

image

Reinmar:
Agree with @oleq. This fix should land in cke5-widget.

pomek:
Correct PR will land to the widget repository.

PR #287, opened 2019-04-17
Fix "_rwd.css" import path

pomek:

Other: Aligned import paths after moving `_rwd.css` file from `theme-lark` to `ui`. See ckeditor/ckeditor5#1662.


pomek:
Requires: https://github.com/ckeditor/ckeditor5-ui/pull/484

PR #297, opened 2019-07-12
Image widgets should not be forced to consume an entire width

### LINKED ISSUE #1870: Image widgets should not be forced to consume an entire width
## Is this a bug report or feature request?

🆕 Feature request

## 💻 Version of CKEditor

CKEditor v5 @ 12.1.0

## 📋 Steps to reproduce

Image widgets (the wrapper) currently consume entire contianer's width, even if an image is much narrower than the container.

## ✅ Expected result

Wrapper should be reduced to minimal required width.

## ❎ Actual result

image

Wrapper spans over entire container's width.

dkonopka:

Other: Image widgets should not the entire width of the editor. Closes https://github.com/ckeditor/ckeditor5/issues/1870.

BREAKING CHANGE: From now on, all images use CSS `display: table` by default (`.ck-content .image { display: table }`). This could have an impact on your application and we recommend you to check if images render correctly in CKEditor 5 after this update.

---

### Additional information

Screenshot 2019-07-12 at 13 25 31

### DIFF
--- a/theme/image.css
+++ b/theme/image.css
@@ -4,11 +4,12 @@
  */
 
 .ck-content .image {
+	display: block;
 	clear: both;
 	text-align: center;
 
-	/* Make sure there is some space between the content and the image. */
-	margin: 1em 0;
+	/* Make sure there is some space between the content and the image. Center image by default. */
+	margin: 1em auto;
 
 	& > img {
 		/* Prevent unnecessary margins caused by line-height (see #44). */
--- a/theme/image.css
+++ b/theme/image.css
@@ -20,5 +20,8 @@
 
 		/* Make sure the image never exceeds the size of the parent container (ckeditor/ckeditor5-ui#67). */
 		max-width: 100%;
+
+		/* Make sure the caption will be displayed properly (See: https://github.com/ckeditor/ckeditor5/issues/1870). */
+		min-width: 50px;
 	}
 }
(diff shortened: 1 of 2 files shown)


oleq (commented):
There's a brief moment the caption becomes vertical, e.g. when using images from CKFinder

Kapture 2019-07-17 at 17 33 55

Can we do anything about that?

Reinmar:
> Can we do anything about that?

Min-width for `<figure>`? Something like 50px?

Reinmar:
I mean – it seems silly – like a hack, not a solution. But:

* in 99.9% cases you won't be inserting image smaller than that anyway,
* even if you do, you won't be able to have a caption inside

Also, another thing – we should find a CSS setting which doesn't split a single word such as "caption" into multiple lines. That will stretch the thing itself too.

Reinmar:
3rd comment – 3rd thought

> Also, another thing – we should find a CSS setting which doesn't split a single word such as "caption" into multiple lines. That will stretch the thing itself too.

OTOH, AFAIR we tuned some CSS due to too long words being used in captions (such as 50 x "a" characters). I think this is kinda edge-casey. Buuuut, assuming that you can set the width of the image, you can set it to something like 50px and then long words indeed may need to be wrapped 🤦‍♂ Or not?

oleq:
@dkonopka Can you give it a try? A sane `min-width` sounds like a solution. No one is going to upload a 30px image.

PR #311, opened 2019-08-20
Workaround for Safari's image size bug

### LINKED ISSUE #1975: [Safari] Images uploads with incorrect size
## Is this a bug report or feature request? (choose one)

🐞 Bug report

## 💻 Version of CKEditor

Latest master

## 📋 Steps to reproduce

1. Upload some bigger image.
2. If the issue doesn't occur, repeat step 1 few times.

## ✅ Expected result
The image is uploaded with a correct size.

## ❎ Actual result
The image is uploaded with incorrect size.

## 📃 Other details that might be useful

**Screenshot**
Screenshot 2019-08-20 at 08 17 06

Tested on Safari 12.1.

Reinmar:

Fix: Worked around Safari's image size bug. Closes ckeditor/ckeditor5#1975.

### DIFF
--- a/src/imageupload/imageuploadediting.js
+++ b/src/imageupload/imageuploadediting.js
@@ -197,7 +197,18 @@ export default class ImageUploadEditing extends Plugin {
 				// https://github.com/ckeditor/ckeditor5/issues/1975
 				if ( env.isSafari ) {
 					editor.ui.once( 'update', () => {
+						// Early returns just to be safe. There might be some code ran
+						// in between the outer scope and this callback.
+						if ( !viewImg.parent ) {
+							return;
+						}
+
 						const domFigure = editor.editing.view.domConverter.viewToDom( viewImg.parent );
+
+						if ( !domFigure ) {
+							return;
+						}
+
 						const originalDisplay = domFigure.style.display;
 
 						domFigure.style.display = 'none';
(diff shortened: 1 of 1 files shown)


Reinmar:
Waiting for the CI and if it's ok, we can merge this, I think.

src/imageupload/imageuploadediting.js
```diff
@@ -192,6 +193,33 @@ export default class ImageUploadEditing extends Plugin {
...
+				/* istanbul ignore next */
+				if ( env.isSafari ) {
+					editor.editing.view.once( 'render', () => {
+						// Early returns just to be safe. There might be some code ran
+						// in between the outer scope and this callback.
+						if ( !viewImg.parent ) {
+							return;
+						}
+
+						const domFigure = editor.editing.view.domConverter.mapDomToView( viewImg.parent );
```
Reinmar: I'm stupid. Wrong direction.

src/imageupload/imageuploadediting.js
```diff
@@ -192,6 +193,34 @@ export default class ImageUploadEditing extends Plugin {
...
+						if ( !domFigure ) {
+							return;
+						}
+
+						const originalDisplay = domFigure.style.display;
+
+						domFigure.style.display = 'none';
+
+						// Make sure this line will never be removed during minification for having "no effect".
+						domFigure._ckHack = domFigure.offsetHeight;
```
Reinmar, line 218: Comment for a future me – this is purely theoretical case today that this may be removed during minification. That'd have to be a really eager minifier which would be removing things that don't have any side effect. Perhaps even impossible to use in practice today due to code which actually use such constructs to leave some effects. But I imagined my future self trying to debug why this hack for Safari doesn't work in a minified code only and with this specific minifier and I thought – noooo, I won't do that to myself :D

PR #326, opened 2019-09-02
Make Clipboard plugin required by ImageUploadEditing.

### LINKED ISSUE #2910: Get rid of soft requirements
Background: https://github.com/ckeditor/ckeditor5-core/issues/192 and https://github.com/ckeditor/ckeditor5-core/issues/148.

There are two types of plugins which we're not currently listing in `requires()`:

* Optional ones – those which may change the logic of your plugin when they are loaded, but which are not necessary for your plugin to work.

    Example: if `Notification` is loaded, use its API to display a warning. Do nothing otherwise.

* Soft (dynamic) dependencies – those which are in fact required for your plugin to work correctly, but we didn't want to specify which implementation of this functionality we depend on. Instead, a bit like in "program to an interface" we wanted to depend on an interface. The fact that a plugin called `'Clipboard'` will be available, but we don't need to assume which instance it is precisely.

The first mechanism works fine. It was improved in https://github.com/ckeditor/ckeditor5-core/issues/148 (`plugins.get()` throws, so you get the error ASAP). 

The second might be improved (with soft requirements – using strings in `requires()` as initially discussed in https://github.com/ckeditor/ckeditor5-core/issues/148), but it's actually not used by us. I checked @jodator's research (https://github.com/ckeditor/ckeditor5-core/issues/148#issuecomment-445794316) and did some more and we use this approach in very few cases.

Therefore, let's agree that we don't use soft requirements. If you depending on a clipboard, define it by a class in your requirements. This isn't ideal, but this is how reality works, so I have no intention to deny it that :D.

Also, there's always the option to use [this technique](https://ckeditor.com/docs/ckeditor5/latest/builds/guides/faq.html#how-to-customize-the-ckeditor-5-icons) with [`NormalModuleReplacementPlugin`](https://webpack.js.org/plugins/normal-module-replacement-plugin/) to inject a different implementation by tuning webpack's configuration. While less convenient, the option exists, so no one will be locked.

## What needs to be done?

We need to use `requires()` in places where it's missing now – this can be found by scanning for `plugins.get( '` and filtering out the "optional dependency" cases.

jodator:

Other: Make Clipboard plugin required dependency of ImageUploadEding. Closes ckeditor/ckeditor5#2910.

---

### Additional information

* This is the only required fix I've found for 'soft' requires in CKEditor5 code base.
* ps.: I didn't add the `BREAKING CHANGE:` entry as I don't see this a breaking change but might be needed from a puristic POV: 
```
BREAKING CHANGE: The ImageUploadEding now requires Clipboard plugin to work.
```
* also the `BallonToolbar` usage is OK:

> About that `BalloonToolbar` plugin use - it is OK. It is not a plugin but a class of and BalloonEditor which always requires `BalloonToolbar`.

### DIFF
--- a/src/imageupload/imageuploadediting.js
+++ b/src/imageupload/imageuploadediting.js
@@ -10,6 +10,7 @@
 import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
 import FileRepository from '@ckeditor/ckeditor5-upload/src/filerepository';
 import Notification from '@ckeditor/ckeditor5-ui/src/notification/notification';
+import Clipboard from '@ckeditor/ckeditor5-clipboard/src/clipboard';
 import UpcastWriter from '@ckeditor/ckeditor5-engine/src/view/upcastwriter';
 import env from '@ckeditor/ckeditor5-utils/src/env';
 
@@ -27,7 +28,7 @@ export default class ImageUploadEditing extends Plugin {
 	 * @inheritDoc
 	 */
 	static get requires() {
-		return [ FileRepository, Notification ];
+		return [ FileRepository, Notification, Clipboard ];
 	}
 
 	static get pluginName() {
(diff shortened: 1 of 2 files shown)


Reinmar (changes requested):
Some issue with coverage.

jodator:
It is because of number of lines have decreased and we still didn't fix coverage here.

src/imageupload/imageuploadediting.js
```diff
@@ -118,31 +119,29 @@ export default class ImageUploadEditing extends Plugin {
...
-		if ( editor.plugins.has( 'Clipboard' ) ) {
-			this.listenTo( editor.plugins.get( 'Clipboard' ), 'inputTransformation', ( evt, data ) => {
-				const fetchableImages = Array.from( editor.editing.view.createRangeIn( data.content ) )
-					.filter( value => isLocalImage( value.item ) && !value.item.getAttribute( 'uploadProcessed' ) )
-					.map( value => { return { promise: fetchLocalImage( value.item ), imageElement: value.item }; } );
-
-				if ( !fetchableImages.length ) {
-					return;
-				}
+		this.listenTo( editor.plugins.get( 'Clipboard' ), 'inputTransformation', ( evt, data ) => {
```
Reinmar:
```suggestion
		this.listenTo( editor.plugins.get( Clipboard ), 'inputTransformation', ( evt, data ) => {
```

PR #330, opened 2019-09-10
Improved image resize performance

### LINKED ISSUE #310: Resizer performance
In general, the performance of the resizer is not great. We do simple debouncing, but it does not actually change anything. I think it was added there prematurely.

This is no debouncing of mousemove:

image

And this is the current master:

image

If you look at the frames, you can actually see better results without debouncing. But my guess is – this might've been due to me dragging a bit differently in both cases (different speed). In any case, as long as we get only 5fps, debouncing seems completely useless, if not harmful. I'm not sure about this, but my guess will be that the browser throttles mousemove itself anyway – it'd be nice to confirm that.

To really understand the performance we need to dig into the timeline. To see why we get such long frames. Let's see how long it takes to handle one mousemove.

This is a single task:

image

Let's analyse it:

1. The first peak is `updateSize()` call so the actual mousemove listener. It looks quite ok:

   image

   We can see that there two costly actions – changing the size of the image itself. And then of the wrapper. Unfortunately, between those actions we need to take the rect to learn how the image grew. This is because we want to position the resize wrapper pixel perfect, regardless of the styling. We could perhaps save some time optimising how the resize wrapper is positioned, but the truth is that this isn't crucial in this picture because the entire `updateSize()` function takes just a fraction of the entire task (2ms from 16ms).

2. Then we get to this part:

    image

    And this is where we've got problems. This entire piece of code should not be executed when the user is resizing an image. 

    * There's no need to do anything with the fake selection for sure. I think we may be trying to fix it because mousedown cause the fake selection container to lose focus and the browser to try to move the selection somewhere else. We should find ways to prevent any interaction with the selection on mousedown and mousemove here.
    * Then we try to reposition the toolbar. We can simply hide the toolbar during resizing. That will simply make the problem go away.
    * Finally, there's `redraw()` from the resizer. Which... I totally don't get. Since `updateSize()` resized the wrapper already, why do we need `redraw()`? It's super costly, BTW. I guess it does something more than `updateSize()` but then it should be clarified and cleaned up what does what for what reason.

### LINKED ISSUE #5191: Resizer performance
In general, the performance of the resizer is not great. We do simple debouncing, but it does not actually change anything. I think it was added there prematurely.

This is no debouncing of mousemove:

image

And this is the current master:

image

If you look at the frames, you can actually see better results without debouncing. But my guess is – this might've been due to me dragging a bit differently in both cases (different speed). In any case, as long as we get only 5fps, debouncing seems completely useless, if not harmful. I'm not sure about this, but my guess will be that the browser throttles mousemove itself anyway – it'd be nice to confirm that.

To really understand the performance we need to dig into the timeline. To see why we get such long frames. Let's see how long it takes to handle one mousemove.

This is a single task:

image

Let's analyse it:

1. The first peak is `updateSize()` call so the actual mousemove listener. It looks quite ok:

   image

   We can see that there two costly actions – changing the size of the image itself. And then of the wrapper. Unfortunately, between those actions we need to take the rect to learn how the image grew. This is because we want to position the resize wrapper pixel perfect, regardless of the styling. We could perhaps save some time optimising how the resize wrapper is positioned, but the truth is that this isn't crucial in this picture because the entire `updateSize()` function takes just a fraction of the entire task (2ms from 16ms).

2. Then we get to this part:

    image

    And this is where we've got problems. This entire piece of code should not be executed when the user is resizing an image. 

    * There's no need to do anything with the fake selection for sure. I think we may be trying to fix it because mousedown cause the fake selection container to lose focus and the browser to try to move the selection somewhere else. We should find ways to prevent any interaction with the selection on mousedown and mousemove here.
    * Then we try to reposition the toolbar. We can simply hide the toolbar during resizing. That will simply make the problem go away.
    * Finally, there's `redraw()` from the resizer. Which... I totally don't get. Since `updateSize()` resized the wrapper already, why do we need `redraw()`? It's super costly, BTW. I guess it does something more than `updateSize()` but then it should be clarified and cleaned up what does what for what reason.

mlewand:

Other: Improved image resize performance. Closes ckeditor/ckeditor5#5191.

---

### Additional information

For main PR logic see the widget repo PR ckeditor/ckeditor5-widget#108.

There's also a related ckeditor/ckeditor5-engine#1792 PR which is vital for this enhancement.

### DIFF
--- a/tests/imageresize.js
+++ b/tests/imageresize.js
@@ -690,6 +690,6 @@ describe( 'ImageResize', () => {
 	}
 
 	function getFirstResizer( editor ) {
-		return Array.from( editor.plugins.get( 'WidgetResize' ).resizersByWrapper.values() )[ 0 ];
+		return Array.from( editor.plugins.get( 'WidgetResize' ).resizers.values() )[ 0 ];
 	}
 } );
--- a/tests/imageresize.js
+++ b/tests/imageresize.js
@@ -454,7 +454,7 @@ describe( 'ImageResize', () => {
 
 			editor.commands.get( 'undo' ).execute();
 
-			await wait( 40 );
+			await wait( 160 ); // ui#update event is throttled.
 
 			const resizerWrapper = document.querySelector( '.ck-widget__resizer' );
 			const shadowBoundingRect = resizerWrapper.getBoundingClientRect();
--- a/tests/imageresize.js
+++ b/tests/imageresize.js
@@ -141,7 +141,7 @@ describe( 'ImageResize', () => {
 		it( 'disables the resizer if the command is disabled', () => {
 			setData( editor.model, `<paragraph>foo</paragraph>[<image src="${ IMAGE_SRC_FIXTURE }"></image>]` );
 
-			const resizer = editor.plugins.get( 'WidgetResize' ).resizers[ 0 ];
+			const resizer = getFirstResizer( editor );
 
 			let isEnabled = false;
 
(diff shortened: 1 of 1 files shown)


Reinmar:
I R-ed the PR in ckeditor5-widget, so this gets automatic R- too.

PR #335, opened 2019-11-19
Improved the performance of the ImageLoadObserver

### LINKED ISSUE #5766: ImageLoadObserver performance issues
Extracted from https://github.com/ckeditor/ckeditor5/issues/4504#issuecomment-479443895.

mlewand:

Fix: Improved markup operation performance of the editor with the image plugin enabled. Closes https://github.com/ckeditor/ckeditor5/issues/5766.

---

### Additional information

Do we want to add a test for that? I feel that integration test for this performance improvement would be a nice protection.

---

Example verification benchmarks:

* base (`master` branch): **11.01s** to remove attributes
* this branch: **2.33s** to remove attributes

Note that this was performed on a idling, real environment notebook, so other factor might distort the result. But it's in line with what you feel manually checking it, so it gives the right idea.

Markup used in the editor can be found here: https://gist.github.com/mlewand/109ae7e518cb91d90318c81331b7d37a

The further performance could be addressed by tweaking engine package.

### DIFF
--- a/src/image/imageloadobserver.js
+++ b/src/image/imageloadobserver.js
@@ -25,7 +25,7 @@ export default class ImageLoadObserver extends Observer {
 	 */
 	observe( domRoot ) {
 		this.listenTo( domRoot, 'load', ( event, domEvent ) => {
-			const domElement = domEvent.path[ 0 ];
+			const domElement = domEvent.target;
 
 			if ( domElement.tagName == 'IMG' ) {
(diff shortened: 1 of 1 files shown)


Reinmar:
> Do we want to add a test for that? I feel that integration test for this performance improvement would be a nice protection.

It'd be nice but we've got too many problems with those tests. They are quite unstable, especially on CI. So, I'd skip that.

PR #342, opened 2020-02-19
Update logic for get 'img' form 'figure'.

kaluginserg:
Update logic for get `img` form `figure` view element.
Fix strange behaviour with custom extending plugins with downcasting images attributes.
Issue describes here: https://github.com/ckeditor/ckeditor5/issues/6294

Default behaviour for getting img-view element was : `.getChild( 0 )` it is not best when we have some tags before img 'tag'. It can be possible in future...

Suggested behaviour: 
`Array.from( figureView.getChildren() ).find( viewChild => viewChild.is( 'img' ) );`

and encapsulate it to utils function `getImgViewFromFigure()`.

This approach provide ability to extend plugin logic with other custom plugins.

---

### DIFF
--- a/src/image/converters.js
+++ b/src/image/converters.js
@@ -8,6 +8,7 @@
  */
 
 import first from '@ckeditor/ckeditor5-utils/src/first';
+import { getImgViewFromFigure } from './utils';
 
 /**
  * Returns a function that converts the image view representation:
@@ -81,7 +82,7 @@ export function srcsetAttributeConverter() {
 
 		const writer = conversionApi.writer;
 		const figure = conversionApi.mapper.toViewElement( data.item );
-		const img = figure.getChild( 0 );
+		const img = getImgViewFromFigure( figure );
 
 		if ( data.attributeNewValue === null ) {
 			const srcset = data.attributeOldValue;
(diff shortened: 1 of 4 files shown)


kaluginserg:
@Reinmar @dkonopka 
continuous-integration build failed - but detailed reason seems not related with MergeRequest itself. 
Can you review my Merge Request ?
Thanks.

jodator:
Hello - thanks for the fix we spotted that error but haven't had a use-case for such extension so it was left as-is.

I didn't check if it solves all of the issues with extending the image feature output, but the change itself looks OK.

If you could add a test for this would be awesome. I'll check this next week after the release process and most likely merge it.

Reinmar:
As for the CI, please ignore that. It's something we'll need to resolve.

kaluginserg:
@jodator 
About test cases I have no test cases, I think - needs to create some little extender, or sub-plugin, like current figcaption (but added before img tag)...

I can share with you our company concept for title/legends for images/tables, (it is almost work for images, with little bugs, but not work for tables itself, for tables we use this approach with extra-wrapper, because built-in table - hard extensible):
https://gist.github.com/kaluginserg/497d5702d1444d1284f660ec5cc27083

jodator:
Thank you @kaluginserg for fixing this. I've added one test to this. You can track the next release announcements here: https://github.com/ckeditor/ckeditor5/issues/6317

